PR #118, opened 2014-12-22
Fix of colors so it doesn't affect the reading of bash.

lobocode:
Correcting the symbology of common user "$" when changing to super user "#". The old-mode way, stayed with $ in the two cases. Modified also to users who use the command su to login as user with super user permission in addition to the root itself.


lobocode:
Thank you for giving merge.

2015-04-27 14:00 GMT-03:00 Satyajit Sahoo <notifications@github.com>:

satya164:
Welcome.

Satyajit Sahoo
UX Designer
Behance Profile <https://www.behance.net/satyajitha28c7>

We're all stories, in the end. Just make it a good one, eh? — The Doctor,
Season 5, Episode 13.

Issue #127, opened 2015-02-02
Suggestion: Popcorn Time


Rohan62442:
Popcorn Time is a multi-platform, open source Bittorrent client that also includes a media player. Only instead of actually streaming unauthorized copies of films using Bittorrent, Popcorn Time uses sequential downloading and uploading to play films in its media player.

PR #128, opened 2015-02-02
Added Popcorn Time to additional software.

Rohan62442:
Fixes issue #127.


Rohan62442:
Sorry for the multiple commits....

cadesalaberry:
Because Popcorn Time is installed as `root`, it cannot auto-update itself. I fixed it with
```bash
$ chown $(whoami):$(whoami) /opt/popcorn-time/
```

But I feel like there are more elegant solutions. @satya164 @Rohan62442 any idea ?

Issue #133, opened 2015-02-11
Suggestion : atom


ripper2hl:
fedy supports brackets and sublime but no atom  https://atom.io/

ripper2hl:
Woww!! thanks fedy is a great tool !!

Rohan62442:
This issue can be closed. The pull request (#134) has been merged.

PR #134, opened 2015-02-12
Added Atom text editor to Additional software

Rohan62442:
This solves issue #133 

NOTE: Atom is only available for 64 bit (x86_64) systems. See https://github.com/atom/atom/blob/master/README.md


satya164:
Would be better to hide the entry on 32-bit systems. You can do the following,

```
atom_editor_hide() {
if [[ ! "$arch" = "64" ]]; then
    printf "true"
fi
}
```

Rohan62442:
That's done...

PR #135, opened 2015-02-13
Add support for IntelliJ IDEA, both Community and Ultimate editions

gabrielbauman:
This patch adds support for both editions of IntelliJ IDEA. See http://www.jetbrains.com/idea/

I've tested both plugins and had no issues. If there are any changes you require before merging, let me know.


Rohan62442:
You've used yum to install JDK in the plugin. You can use install_pkg function.
Like this -
install_pkg "java-1.8.0-openjdk" "java-1.8.0-openjdk-devel"

Issue #131, opened 2015-02-08
Install Viber


valentt:
Hi, please add Viber as software install.

Viber is available from rpm - http://download.cdn.viber.com/desktop/Linux/viber.rpm

satya164:
Can you send a pull request? Should be an easy thing to add. Code will very similar to the google chrome installer code.

Rohan62442:
This issue can be closed. The pull request (#140) has been merged.

satya164:
You might wanna check out the `next` branch. There are major changes!

Rohan62442:
I'm doing that right now. How do I pull that branch to my local repo?

satya164:
Just do the following,

```
git pull
git checkout next
./fedy
```

satya164:
I'd suggest to go with the next branch and don't worry about master.

PR #140, opened 2015-02-16
Added Viber to Additional Software

Rohan62442:
This solves issue #131 

The first 4 commits are also included in pull request #134 because I made those changes on the master (which I shouldn't have, so sorry for that). I think merging #134 and then this would solve the problem? If not, @satya164 please let me know.


satya164:
Thanks for your series of pull requests :D

I'll review and merge all in the weekend. Busy with Ozon OS stuff right now.

Satyajit Sahoo
Digital artist
DeviantArt Profile <http://satya164.deviantart.com>

We're all stories, in the end. Just make it a good one, eh? — The Doctor,
Season 5, Episode 13.

PR #156, opened 2015-04-29
I fixed the variable lobo within the EOF.

lobocode:
I fixed the variable $(whoami) within the EOF with backslash bar so it can be read correctly. Now is working correctly ( I tested ) .


lobocode:
I added the chromium-pepper-flash17 (that makes chromium and google chrome compatible with Adobe Flash 17) added a functionality package-cleanup --orphans and a systemd functionality, analyze plot for user analyzing a boot performance , and I fix a small failure doesn't color_prompt.

lobocode:
Beyond all the other made modifications already, I made an modification of folders and architecture of folders. It makes better to separate softwares of development in an only session and additional software in another. I wait that it likes them modifications.

satya164:
This pull request is no longer mergeable. Please rebase against the current master.

Issue #94, opened 2014-05-23
BitTorrent Sync (btsync)


kaveiros:
Could you please consider adding the installation option for btsync ? just a little script that will allow newbies set their password, port for btsync, and add it to systemd without having to "dig" through config files so it will run at startup easily. Thank you in advance.

satya164:
I've been busy lately. So if someone could send a pull request, it'll be great.

Rohan62442:
I'm working on this...

Rohan62442:
I'm having a problem however. It's recommended that btsync not be installed as root, but fedy itself runs with root privileges. So it is a problem. Also, it gives a permission denied error when btsync (installed by fedy) is run... Any help?

satya164:
Can you point me to the branch? I'll investigate.

Rohan62442:
Here -> https://github.com/Rohan62442/fedy/tree/bittorrent-sync

satya164:
@Rohan62442 Any updates on this? I would be great if you could redo the plugin according to the new structure before I release the next version of Fedy.

user501254:
Done in Commit d57ce5ee512bf2db553001175b2a7d46ffd02088.

PR #172, opened 2015-05-24
Added BitTorrent Sync

Rohan62442:
See #94


plugins/bittorrentsync.plugin/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+	"scripts": {
+		"exec": {
+			"label": "Install",
+			"command": "sh install.sh"
+		},
+		"undo": {
+			"label": "Remove",
+			"command": "sh uninstall.sh"
+		},
+		"status": { "command": "bash -c 'test -e ~/.local/share/bittorrent-sync/btsync'" }
```
satya164: You can just use `test -e ~/.local/share/bittorrent-sync/btsync`

satya164:
This installs the app for the current user only, while other plugins install apps for all users. May be we could do that here too?

Fedy provides a helper command `run-as-root`, which is just an alias for `pkexec` to show a GUI password prompt.

Note that `pkexec` doesn't keep the working directory, so when trying to run a script situated in the plugin directory as root, you need to do it like `run-as-root -s install.sh`

Thanks :)

Rohan62442:
If you check the issue I had in the previous versions of Fedy, it was recommended that BitTorrent Sync not be installed as root. Now I no longer remember why... :(

I will check the root installation (I knew about `run-as-root`)

satya164:
I did some search and they say not to run btsync as root. Didn't find anything about installing as root though.

I also came across this wiki - https://wiki.archlinux.org/index.php/BitTorrent_Sync

The wiki also installed it as root.

Rohan62442:
OK, then. I will modify this.

Rohan62442:
I have not tested this though..

plugins/bittorrentsync.plugin/install.sh
```diff
@@ -0,0 +1,47 @@
...
+if [[ ! -f "$FILE" ]]; then
+	exit 1
+fi
+
+mkdir -p "/opt/bittorrent-sync"
+tar -xf "$FILE" -C "/opt/bittorrent-sync"
+ln -sf "/opt/bittorrent-sync/btsync" "/usr/bin/btsync"
+
+btsync --dump-sample-config > /etc/btsync.conf
+systemctl enable btsync
```
satya164: I don't think you can enable or start services without installing them

Rohan62442:
Well, for me, the root install does not work.

Jobava:
Another way to install it may be to create a separate user and group for that particular app with its own permissions and selinux rules, as it's sometimes done for apache.

PR #183, opened 2015-05-30
Added LightTable

Rohan62442:
Added the LightTable IDE/text editor

### DIFF
--- a/plugins/lighttable.plugin/install.sh
+++ b/plugins/lighttable.plugin/install.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/lighttable"
+
+if [[ "$(uname -m)" = "x86_64" ]]; then
+	ARCH="64"
+else
+	ARCH=""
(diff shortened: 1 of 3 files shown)


Rohan62442:
There is currently an open issue with LightTable on Linux 64 bit 

[https://github.com/LightTable/LightTable/issues/1894] (LightTable issue)

It gives the following error:

```ln: cannot remove ‘/opt/LightTable/libudev.so.0’: Permission denied
/opt/LightTable/ltbin: error while loading shared libraries: libudev.so.0: cannot open shared object file: No such file or directory```

It still has a workaround. I ran this on Fedora 21 (64 bit). You need to track down `libudev.so.1` on your system and create a symlink to LightTable's `libudev.so.0`

`sudo ln -sf "/usr/lib64/libudev.so.1" "/opt/LightTable/libudev.so.0"`

satya164:
I needed to make some changes  22d070eb4a26401335b47b06b1c73bb5fb7ec551 .

Issue #180, opened 2015-05-27
Hability to select Chrome version


isantolin:
Its a Good idea in the Chrome install Button, a selector for Beta or Stable version

PR #195, opened 2015-06-06
Added Google Chrome Beta

PhnxRbrn:
Added Google Chrome Beta


satya164:
The pull request is not mergable.

BTW, I'm not really sure about merging this. It might confuse a novice user so as to which version of Chrome to install. Having different versions of Sublime Text, IntelliJ idea etc. are ok, coz they are targeted to advanced users, who will know which one to install.

PhnxRbrn:
OK, that is fine.  I know there was someone asking for chrome beta in issue #180 and I was not sure how to do a selector and thought that maybe this was the next best thing.

PhnxRbrn:
If you want me to fix this to accommodate issue #180, I can.  Otherwise, close this request.

satya164:
@PhnxRbrn No problem. May be I could think of something in the future, like a overflow menu for things like this, which are hidden, but available for advanced users. Let's see.

Pull requests are always welcome though. There are lots of other requests. May be you can have a look if you want to contribute. Thanks.

Issue #179, opened 2015-05-27
Require some more softwares


gauravdighe:
Previous version of Fedy had more softwares.
Anyways I would love to see some more softwares like handbrake-gtk, conky, coverglobus, some more very attractive themes and icons

testman42:
In fact, why not make Fedy distro-agnostic and contact Ninite dudes and propose merge, to make universal popular-software-installer for every OS.

satya164:
@testman42 Fedy core is distro-agnostic. In fact someone is making something for Ubuntu based on Fedy.

Rohan62442:
Conky is already a part of the official fedora repository so it adds no value to Fedy. And Covergloobus has not been in development since 2010.

Handbrake has been added now and Fedy has numix, arc, ozon and evopop themes.

PR #202, opened 2015-06-06
Added HandBrake

Rohan62442:
Added HandBrake-gui and Handbrake-cli, using negativo17 handbrake repository. See #179


satya164:
Uninstalling should also delete the repo.

Thanks :)

Rohan62442:
The Handbrake repo also contains other software like libdvdcss and makemkv. Removing the repo on handbrake's uninstall may cause problems for those people who are using them.

satya164:
libdvdcss is already in the ozon repo. May be detect if the user is using those software and remove the repo if not?

Rohan62442:
Well, it's in this repo as well. So it might give a false positive. Any suggestions?

satya164:
Hey, can you fix the merge conflicts?

PR #205, opened 2015-06-07
Added VisualStudioCode plugin

nmabhinandan:
VSCode is a lightweight cross-platform code editor from Microsoft.
https://code.visualstudio.com/


satya164:
Thanks. But I'll change the command to "vscode" instead of "code", coz it sounds too generic.

nmabhinandan:
:+1:

PR #215, opened 2015-06-20
Fix Popcorn Time Icon

dpocina:
Changed Popcorn Time icon name from popcorn-time to popcorntime.
That is the icon name used by most icon themes (including Ozon).
Before even if you had an icon theme installed you would always see the default icon.


Rohan62442:
Merged after cherry-picking the commit locally.

Issue #211, opened 2015-06-09
Suggestion: clear label with software license


Gwend4l:
Hi!

A quick suggestion: it would be really great to have a clear way to know, for each fedy item, (1) whether or not it's libre software or proprietary software, and (2) the software license.

For instance:

- **Adobe Flash**
    - Browse plug-in and rich internet application runtime.
    - *Proprietary*
- **Atom**
    - A hackable text editor for the 21st Century
    - *Libre (MIT License)*

Or

- **Better font rendering**
    - Improve the font rendering for better readability
    - *Requires patented/proprietary stuff*

Note that for the last example I simply don't know, but I do know that some font rendering can be based an  patented/proprietary stuff (e.G. subpixel rendering is patented), and if it's the case that would be awesome to make it appear. 

My 2 cents! Thanks for Fedy, great work :)

Jobava:
I like this idea too. Would you like to try a pull request?

Gwend4l:
I would, but I am going to be too busy for another week or two. I'll give it a try then, if you're fine with that.

satya164:
We will do more improvements to the license display in future. For now just showing the license should do.

PR #223, opened 2015-06-30
Adding metadata and logic to display software licenses

Gwend4l:
As suggested in issue #211, here is an attempt to add new metadata to describe software licenses, and some code to display such information.

Note that this is a first sketch, I am not sure at all that the fields I've added are the right ones, and anything/everything can be reworked. Don't hesitate to reject all this :)

Regarding **metadata**, here are the fields I've added:

- `license`, a string to describe the license (e.g. GPLv3, Commercial)
- `libre`, a boolean to state that the plugin contains libre software
- `proprietary`, a boolean to state that that this is proprietary software
- `patented`, a boolean to state that it contains patented techniques

All the information was mostly taken from Wikipedia, RPMs metadata and Official web pages.

Regarding the **code**, it impacts the number of rows used by a plugin description, hence I've added an early check to know whether we need 2 or 3 rows (`has_license` boolean). Then to display the license, it works the following way:

- If only the `libre` boolean is true, it displays "Libre"
- If only the `proprietary` boolean is true, it displays "Proprietary"
- If only the `patented` boolean is true, it displays "Patented"
- In other cases, it displays "Mixed"

Then the license description is given in between parentheses.

Here is what it currently looks like:
capture d ecran de 2015-06-30 16-13-23

### DIFF
--- a/app.js
+++ b/app.js
@@ -472,6 +472,8 @@ const Application = new Lang.Class({
             for (let item in this._plugins[category]) {
                 let plugin = this._plugins[category][item];
 
+                let has_license = plugin.libre || plugin.proprietary
+
                 let grid = new Gtk.Grid({
                     row_spacing: 5,
                     column_spacing: 10,
@@ -516,7 +518,13 @@ const Application = new Lang.Class({
                     image.set_from_icon_name("system-run", Gtk.IconSize.DIALOG);
                 }
 
-                grid.attach(image, 0, 1, 1, 2);
+                let nb_rows
+                if (has_license) {
+                  nb_rows = 3
+                } else {
+                  nb_rows = 2
+                }
+                grid.attach(image, 0, 1, 1, nb_rows);
 
                 let label = new Gtk.Label({
                     halign: Gtk.Align.START,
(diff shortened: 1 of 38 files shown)


Rohan62442:
Instead of having multiple booleans, wouldn't it be better to have a single string, say *license*, which will display the text?

Gwend4l:
I started by doing that, but then I realised that there are lots of terms used to talk about FOSS software (e.g. free, open source), and that this could lead to something less homogeneous than controlling what to display in the code (e.g. displaying "libre"). It's also a way to change our mind and to decide that "Proprietary" is not the right word, but "Freeware" is (this is just an example), and to change this without editing any json file.

In other words, it makes sure that everything is formatted in a regular way such as:
[Libre|Proprietary|Mixed] (<description>)

Yet you're probably right; in fact I did begun with a simple string. It might still be the best option to have less lines of code and less metadata fields than with such booleans. If you prefer, I can roll back to this simpler solution.

satya164:
Thank you for your effort. I'll try to visually improve it sometime.

PR #230, opened 2015-07-10
Added Mumble

PhnxRbrn:
Added Mumble from copr.  Fedora 22 from jamielennox and Fedora 21 from lkiesow.


satya164:
If it's only for 64-bit, you should hide the plugin under 32-bit. See here - https://github.com/folkswithhats/fedy/blob/master/plugins/grub2.plugin/metadata.json#L10

PhnxRbrn:
I understand I need to add the show line but how do I specify that it only shows for 64bit

satya164:
@PhnxRbrn Something like following should work,

"show": { "command": " test $(uname -m) = x86_64" }

PhnxRbrn:
OK I will update both mumble and obs-studio to fox this

PhnxRbrn:
Updated

Rohan62442:
@PhnxRbrn Could you rework this using lkiesow's [Copr] (https://copr.fedoraproject.org/coprs/lkiesow/mumble/) repo and submit it again? I believe it includes all Fedora versions now..

PR #246, opened 2015-07-30
Added PyCharm Community Edition

Rohan62442:
PyCharm Community Edition via Copr


satya164:
I'm thinking of moving all these IDE plugins to a separate repo and add a new package `fedy-plugins-ide` or something, coz there are just too many of them.

Rohan62442:
Agreed.

satya164:
@Rohan62442 Yes, but I'll keep Sublime Text 3, since I've seen lots of non-devs use it.

PR #248, opened 2015-07-30
Using rommon/telegram copr repo.

Fnux:
I think it would be better to use this repo : https://copr.fedoraproject.org/coprs/rommon/telegram/
It was cited few weeks ago in FedoraMagazine : http://fedoramagazine.org/telegram-in-fedora/


PhnxRbrn:
The problem I have had with this repo is that since this application has an auto update feature, if it auto updates before the repo updates, you end up with two versions of telegram.

PhnxRbrn:
This is actually also mentioned in that same article in FedoraMagazine.

PR #253, opened 2015-08-22
Update for install.sh for Telegram indicator

thunderbirdtr:
https://github.com/telegramdesktop/tdesktop/issues/491    (last 3 comments) 

According the "Telegram" issues list in GNOME desktop we gonna need "libappindicator" for systray option. Should we add this too in install or just make separation for it (GNOME only, Other DE(s) ) I don't know but we need this lib in order to make it work.

Thank you

### DIFF
--- a/plugins/telegram.plugin/install.sh
+++ b/plugins/telegram.plugin/install.sh
@@ -2,4 +2,4 @@
 
 dnf copr -y enable rommon/telegram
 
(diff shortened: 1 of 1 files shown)


satya164:
I guess it should be okay to install it by default.

thunderbirdtr:
Thank you :+1:

PR #258, opened 2015-09-20
Updated chrome install.sh

DStewart1997:
Updated chrome install.sh to include the fix to prevent the duplicate icons in docks.
Fix found here - http://kb.openstudioproject.com/content/fix-double-google-chrome-icon-docky-and-plank


satya164:
This pull request has a lot of commits. Can you cherry-pick the specific commit to new pull request?

DStewart1997:
Sorry don't know how that happened with that. Will deal with it when I get home from work. 

What happened to the other commits? Have you removed them? Eg. Spotify & PlayOnLinux are no longer there but were when I cloned it last week?

satya164:
@DStewart1997 They were some reorganizations and removals.

PR #282, opened 2015-10-16
added a plugin to install android studio 1.4

FrLallemand:
There was a plugin to that before, but it seems it has been removed.


Abhinav1217:
It has not been removed but moved to a different plugin repository. You can download the plugin from there and use it. https://github.com/folkswithhats/fedy-plugins-developers

FrLallemand:
Thanks, I didn't know of this second repository.

Abhinav1217:
It was created to separate Fedy related pull requests from Plugins related pull request.  We are reworking on plugins feature so it would be much more easier in future to look for plugins.

PR #292, opened 2015-11-04
Fixed Status issue

Abhinav1217:
Status now checks for vivaldi-beta instead of preview. Now status is properly checked.


satya164:
Need to fix this as well - https://github.com/folkswithhats/fedy/blob/master/plugins/vivaldi.plugin/metadata.json#L14

Abhinav1217:
Done thanks.

PR #301, opened 2015-11-11
Add Arc Theme Solid

user501254:
A non-transparent version of the Arc theme.

I think a good number of people (including me) like opaque themes.

See:
https://github.com/horst3180/Arc-theme#packages
https://github.com/horst3180/Arc-theme/issues/216


Abhinav1217:
I pushed an update today before looking at this.  I will push this for next update.

user501254:
Thank you.

PR #302, opened 2015-11-13
Add BitTorrent-Sync

user501254:
Not entirely sure if I have taken all test cases into account.

### DIFF
--- a/plugins/bittorrentsync.plugin/install.sh
+++ b/plugins/bittorrentsync.plugin/install.sh
@@ -13,6 +13,6 @@ wget -P /var/tmp/ https://download-cdn.getsync.com/stable/linux-$ARCH/BitTorrent
 mkdir -pv /opt/btsync
 tar xzfv /var/tmp/BitTorrent-Sync_$ARCH.tar.gz -C /opt/btsync
 
-ln -sf /opt/btsync/btsync /usr/bin/btsync
+ln -sf /opt/btsync/btsync /usr/local/bin/btsync
 
 echo "use \"sudo btsync\" to start"
--- a/plugins/bittorrentsync.plugin/install.sh
+++ b/plugins/bittorrentsync.plugin/install.sh
@@ -6,7 +6,9 @@ else
 	ARCH="i386"
 fi
 
-rm -f BitTorrent-Sync_*.tar.gz*
+CACHEDIR="/var/cache/fedy/bittorrent-syn";
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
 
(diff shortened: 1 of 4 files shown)


plugins/bittorrentsync.plugin/install.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [[ "$(uname -m)" = "x86_64" ]]; then
+	ARCH="x64"
+else
+	ARCH="i386"
+fi
+
+rm -f BitTorrent-Sync_*.tar.gz*
+
+wget -P /var/tmp/ https://download-cdn.getsync.com/stable/linux-$ARCH/BitTorrent-Sync_$ARCH.tar.gz
```
satya164: For downloads, follow the same approach as https://github.com/folkswithhats/fedy-plugins-developers/blob/master/intellijideacommunity.plugin/install.sh#L5

.i.e - downloading to a cache dir, and change the wget command to be like wget -c "$URL" -O "$FILE"

plugins/bittorrentsync.plugin/install.sh
```diff
@@ -0,0 +1,18 @@
...
+
+mkdir -pv /opt/btsync
+tar xzfv /var/tmp/BitTorrent-Sync_$ARCH.tar.gz -C /opt/btsync
+
+ln -sf /opt/btsync/btsync /usr/local/bin/btsync
+
+echo "use \"sudo btsync\" to start"
```
satya164: Remove it as it will not be visible to the user.

plugins/bittorrentsync.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pkill btsync
```
satya164: Is this necessary?

plugins/bittorrentsync.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
...
+
+sudo rm -fr /opt/btsync
```
satya164: Remove sudo

plugins/bittorrentsync.plugin/uninstall.sh
user501254: There is no user interface for this program on Linux, except for the webpage that can be viewed at `localhost`. I figured it would be safer to explicitly kill the program so that synchronization of files stops right away if the daemon is running in background.

satya164: It'll be preferable to run something like `btsync stop` if there is such a command, rather than killing the process. Change it if it's possible. Otherwise it's okay.

PR #317, opened 2015-12-25
update to get latest build of Sublime Text 3

user501254:
The old link would always download `build 3083` of Sublime Text 3; ie. file `sublime_text_3_build_3083_x32.tar.bz2` or  `sublime_text_3_build_3083_x64.tar.bz2` even if newer builds are available which is a little undesirable. 

The changes makes it possible to get the most current build of ST3 available on their website.


Abhinav1217:
please add some description on "what's" and "why's" about Commit.

PR #318, opened 2015-12-26
use new rpm package version

user501254:


Abhinav1217:
Getting 503 error in my tests. Also please add some description on "what's" and "why's" about changes.

Issue #295, opened 2015-11-06
Swap NOOP SSD tweak for DEADLINE


egriffith:
Given OpenSUSE's comments in their email to Michael over at Phoronix (article available here: http://www.phoronix.com/scan.php?page=news_item&px=openSUSE-42.1-Leap-Kernel) 

I believe that Fedy should at least consider swapping the NOOP scheduler tweak over to DEADLINE.

Abhinav1217:
We will look into it. Also if you have time, you could give us a pull request.

Abhinav1217:
Issue solved with PR #319

PR #319, opened 2015-12-27
Use deadline for SSDs and HDDs

rishadafc:
In response to #295. 

According to the [Debian wiki](https://wiki.debian.org/SSDOptimization#Low-Latency_IO-Scheduler):
> The dumb "noop" scheduler may be a little faster in benchmarks that max out the throughput, but this scheduler causes noticeable delays for other tasks while large file transfers are in progress.

### DIFF
--- a/plugins/ioscheduler.plugin/metadata.json
+++ b/plugins/ioscheduler.plugin/metadata.json
@@ -1,7 +1,7 @@
 {
 	"icon": "drive-harddisk",
-	"label": "SSD I/O scheduler",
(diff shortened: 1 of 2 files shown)


Abhinav1217:
@rishadafc Take a look at #323 . Also use http://www.shellcheck.net/ tool to check your bash scripts. There are several outdated usage in your scripts. Please Update your plugin.

rishadafc:
@Abhinav1217 Thanks for letting me know and providing the ShellCheck recommendation. I have updated the script and tested it to ensure it works properly.

PR #327, opened 2016-01-12
New Flash Player installer

synxbr:
I wrote a new installer for Adobe Flash Player. It will allow users to use the latest version of Flash Player on Firefox. 

Steps:

1. Check if Google Chrome is installed. 
1.1 If yes, the script will run the other steps.
1.2 If not, it will download and extract Flash Player.

2. Download and compile Fresh Player, a plugin wrapper for Firefox and other NPAPI-based browsers.

3. Install libfreshwrapper-flashplayer.so

Done. The latest Flash Player on Firefox :) 

Also, I renamed the older installer to legacy, so users may choose which one to install (latest or legacy).


Abhinav1217:
I would prefer to keep legacy flash player as it is, and have this as a separate plugin. Because both are technically different.  Pepper flash on firefox also tends to be buggy sometimes. I have faced lots of issue with it in past before moving to pipelight-plugin. Legacy flash (even though really outdated) doesn't crash browser. 

I will need discuss it with rest of the guys on what should we include. This one, Legacy one, or pipelight. In any case, having this as a separate plugin would be a better approach than over-riding the legacy one.

satya164:
Plugin names should be nicer. If we ever support usage from CLI, `flash11` or `flashnew` won't mean much. How about `flashlegacy` and `pepperflash`?

Abhinav1217:
Agreed,  Plugin name like `flash-legacy` for current plugin, and `flash-pepper` for new one would be much more appropriate.  

I am still sceptical about  pepperflash.  I tried this plugin, and it caused my firefox to crash on any page with flash on it.   I reverted back to pipelight.  Thats why I haven't merged this pull yet.

bddenhartog:
@Abhinav1217 @satya164 do we want to close this?

Abhinav1217:
Closed as it is not required anymore. Adobe has decided to once again provide flash plugins for linux.

Issue #328, opened 2016-01-13
"Microsoft TrueType core fonts(install) failed" error


oliverfrost:
*Steps to reproduce:*

1. Install Fedy with script mentoned in redme: http://folkswithhats.org/fedy-installer
2. Launch Fedy
3. On "Apps"  tab click "Install" near "Microsoft TrueType core fonts".
4. Enter sudo password if requested.
5. Wait for installation to end. 

*As actual:* "Microsoft TrueType core fonts(install) failed" error appears. Package is not installed. See logs below.
*As expected:* Package installed with no errors.

*Logs:*
[oliverfrost@localhost ~]$ fedy
Last metadata expiration check performed 0:28:09 ago on Wed Jan 13 22:33:05 2016.
No package $(wget available.
No package http://mscorefonts2.sourceforge.net/ available.
No package -O available.
No package - available.
No package | available.
Package grep-2.22-5.fc23.x86_64 is already installed, skipping.
No package -Po available.
No package https://downloads.sourceforge.net/project/mscorefonts2/rpms/msttcore-fonts-installer-[0-9.]{3}-[0-9].noarch.rpm) available.
Error: Unable to find a match.

*System info:*
Fedora 23 Workstation

[oliverfrost@localhost ~]$ uname -a
Linux localhost.localdomain 4.3.3-300.fc23.x86_64 #1 SMP Tue Jan 5 23:31:01 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

BastienBalaud:
I have a similar problem because "Microsoft TrueType core fonts" need some dependency to install the font and fedy don't install this dependency.
Try to install this package to solve the problem :
curl cabextract xorg-x11-font-utils fontconfig

PR #329, opened 2016-01-14
Fix Microsoft TrueType core fonts plugin

user501254:
- Move commands to separate install and uninstall scripts
- Install dependencies before installing RPM package

Should close https://github.com/folkswithhats/fedy/issues/328.


Abhinav1217:
Thanks for update, I will push an update in few days.

fnkr:
Works for me, thanks!

PR #338, opened 2016-02-09
Git support for Bash color profile.

gregharvey:
Hi,

I've added a little bash profile script I've used for a few years now to this, wrapped in a check for the existence of Git. I think it should be safe, if Git exists and if the current directory is part of a Git repository, it appends Git status and branch information to the prompt. Hope you like this little addition! :-)

Thanks,

Greg

### DIFF
--- a/plugins/colorprompt.plugin/setup.sh
+++ b/plugins/colorprompt.plugin/setup.sh
@@ -1,12 +1,39 @@
 #!/bin/bash
 
 cat <<EOF | tee /etc/profile.d/color_prompt.sh > /dev/null 2>&1
-# Colors in Terminal (Bash)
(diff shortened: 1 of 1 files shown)


satya164:
Here is what I use.

```sh
export PS1="\[\033[36m\]➜\[\033[m\] \[\033[38;5;78m\]\W\[\033[m\]\[\033[38;5;221m\]\$(__git_ps1)\[\033[m\] "
```

Abhinav1217:
Looks like `__git_ps1` was not available in the git-completion file on my system.

Abhinav1217:
Closed as it is not required anymore. F25 onwards use powerline as default.

PR #348, opened 2016-03-21
Added Spacemacs plugin

footvaalvica:
NOTE: This only works if you already have Emacs installed


antmak:
"label": "Revert",
			"command": "run-as-root -s **REVERT**.sh"

footvaalvica:
Thanks for spotting that. Fixed.

Abhinav1217:
I have been using for nearly half a month, It is good editor. It is better suited if you would post the PR to https://github.com/folkswithhats/fedy-plugins-developers . It is where we currently keep all the plugins, separated from fedy app. I would merge PR on that repo. 

I think it would be better if you could add `dnf -y install emacs` on the script itself.  A plugin to install software should install its dependencies as well.  That said, is there is a copr repo for this, it would be great, since it would be easier to manage all its dependencies as  well.

footvaalvica:
I have made a PR on https://github.com/folkswithhats/fedy-plugins-developers. Closing this one.

PR #352, opened 2016-03-29
I added a VirtualMachine to fedy . Hope you like it.

gui1612:


Abhinav1217:
Thanks.

satya164:
VirtualBox is already in the RPMFusion repos. I don't think we need this plugin in core.

PR #353, opened 2016-04-08
Add the mouse plugin

Moop1e:
Sometime useful to disable the mouse acceleration for precision works or games ;)


plugins/mouse.plugin/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+	"label": "Disabling mouse acceleration",
```
satya164: The label should read like `This plugin will **Disable mouse acceleration**`

plugins/mouse.plugin/metadata.json
```diff
@@ -0,0 +1,16 @@
...
+	"description": "Get better mouse precision.",
```
satya164: Can we improve the description please, it's confusing right now.

plugins/mouse.plugin/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cat > /etc/X11/xorg.conf.d/50-mouse-acceleration.conf <<EOF
+Section "InputClass"
+	Identifier "Fedy Mouse Configuration"
```
satya164: This should be more descriptive. `Fedy Mouse Configuration` doesn't tell me anything about what this does.

Moop1e:
Ok, understood. I'll fix it.

satya164:
cc @Abhinav1217

PR #355, opened 2016-04-12
Add Node.js plugin

orlandoxavier:
Hello!
I have added Node.js installation support. Can you accept this pull request?
Regards.


Abhinav1217:
I don't think this is needed. Node.js is included in default repos and several tools like NVM also provide single command for its installation.

PR #374, opened 2016-06-27
added files for gnome titlebar fix plugin

hispeedzintarwebz:
Plugin to remove padding from gnome's overly thick titlebars.  Creates a css file in the user's home directory which overrides system-wide settings, so can be used on a per-user basis.  I still need to find a better way of reverting the changes created, in case the user has already added their own css overrides, as of right now the "undo" simply deletes the file in ~/.config/gtk-3.0/

### DIFF
--- a/plugins/gnometitle.plugin/metadata.json
+++ b/plugins/gnometitle.plugin/metadata.json
@@ -10,7 +10,8 @@
 		},
 		"undo": {
 			"label": "Revert",
-			"command": "rm -f ~/.config/gtk-3.0/gtk.css"
+			"command": "sh ./revert.sh"
 		}
+		"status": { "command": "test if grep -q addedbyfedy ~/.config/gtk-3.0/gtk.css"  }
 	}
 }
--- a/plugins/gnometitle.plugin/metadata.json
+++ b/plugins/gnometitle.plugin/metadata.json
@@ -1,0 +1,16 @@
+{
+	"icon": "utilities-terminal",
+	"label": "Decrease Gnome Title Bar Size",
+	"description": "Decreases thickness of Gnome's default window title bars for the active user.",
+	"category": "Tweaks",
+	"scripts": {
+		"exec": {
+			"label": "Setup",
+			"command": "sh ./setup.sh"
+		},
+		"undo": {
+			"label": "Revert",
+			"command": "rm -f ~/.config/gtk-3.0/gtk.css"
+		}
+	}
+}
--- a/plugins/gnometitle.plugin/setup.sh
+++ b/plugins/gnometitle.plugin/setup.sh
@@ -5,21 +5,26 @@ touch ~/.config/gtk-3.0/gtk.css
 
 #append css overrides
 cat <<EOF >> ~/.config/gtk-3.0/gtk.css
-.header-bar.default-decoration {
-        padding-top: 0px;
-        padding-bottom: 0px;
-    }
-
-.header-bar.default-decoration .button.titlebutton {
-    padding-top: 0px;
-    padding-bottom: 0px;
-}
-
-/* No line below the title bar */
-.ssd .titlebar {
-    border-width: 0;
-    box-shadow: none;
-}
+headerbar { /*addedbyfedy*/
+    min-height: 38px; /*addedbyfedy*/
+    padding-left: 2px; /*addedbyfedy*/
+    padding-right: 2px; /*addedbyfedy*/
+} /*addedbyfedy*/
+headerbar entry, /*addedbyfedy*/
+headerbar spinbutton, /*addedbyfedy*/
+headerbar button, /*addedbyfedy*/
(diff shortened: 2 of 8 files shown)


satya164:
Not sure this is something which should be in core. This is mostly a personal preference.

plugins/gnometitle.plugin/metadata.json
```diff
@@ -0,0 +1,16 @@
...
+	"description": "Decreases thickness of Gnome's default window title bars for the active user.",
+	"category": "Tweaks",
+	"scripts": {
+		"exec": {
+			"label": "Setup",
+			"command": "sh ./setup.sh"
+		},
+		"undo": {
+			"label": "Revert",
+			"command": "rm -f ~/.config/gtk-3.0/gtk.css"
```
satya164: This will remove any other customisations the user has made.

plugins/gnometitle.plugin/metadata.json
```diff
@@ -0,0 +1,16 @@
...
+	"icon": "utilities-terminal",
+	"label": "Decrease Gnome Title Bar Size",
+	"description": "Decreases thickness of Gnome's default window title bars for the active user.",
+	"category": "Tweaks",
+	"scripts": {
+		"exec": {
+			"label": "Setup",
+			"command": "sh ./setup.sh"
+		},
+		"undo": {
```
satya164: `undo` won't be visible without a way to know the current status

plugins/gnometitle.plugin/metadata.json
hispeedzintarwebz: Added revert.sh to remove lines tagged as added by fedy.  Working on status section of metadata now.

plugins/gnometitle.plugin/metadata.json
hispeedzintarwebz: Working on this now

bddenhartog:
@hispeedzintarwebz Might I recommend restructuring your files to be a bit less verbose? Try this out:

`setup.sh`:

```shell
#!/bin/bash

# create css file
touch ~/.config/gtk-3.0/gtk.css

#append css overrides with tags for removal later without removing entire file
cat <<EOF >> ~/.config/gtk-3.0/gtk.css

/* {{{ begin fedy-gtkhidpi }}} */

headerbar {
    min-height: 38px;
    padding-left: 2px;
    padding-right: 2px;
}

headerbar entry,
headerbar spinbutton,
headerbar button,
headerbar separator {
    margin-top: 2px;
    margin-bottom: 2px;
}

.default-decoration {
    min-height: 0;
    padding: 2px
}

.default-decoration .titlebutton {
    min-height: 26px;
    min-width: 26px;
}

/* {{{ end fedy-gtkhidpi }}} */

EOF

#restart gnome shell
gnome-shell -r &
disown
```

To remove the relevant lines from `~/.config/gtk-3.0/gtk.css`, all you need to do now is:

```bash
sed '/begin fedy-gtkhidpi/,/end fedy-gtkhidpi/d' ~/.config/gtk-3.0/gtk.css | sed '/^$/N;/^\n$/D'
```

This simply makes the css file a bit more readable if the someone happened to open it up directly after running your script :)

bddenhartog:
@satya164 Maybe under the `Tweaks` category? We have a mixture of personal preference plugins there (system-wide touchpad click, deadline scheduler, bash colors). Although since this is specifically for the Gnome DE, I can see the argument for not including it.

@hispeedzintarwebz Do you have any plans to make those changes? If not I can edit this directly.

plugins/gnometitle.plugin/setup.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# create css file
+touch ~/.config/gtk-3.0/gtk.css
+
+#append css overrides with tags for removal later without removing entire file
+cat <<EOF >> ~/.config/gtk-3.0/gtk.css
```
bddenhartog: You can simplify this to be a lot more concise. Please [see this comment](https://github.com/folkswithhats/fedy/pull/374#issuecomment-259332593) under the "Discussion" tab for this pull request.

hispeedzintarwebz: Got a new job and haven't revisited this in a while.  Feel free to edit, if not I will try to look at this this weekend.  Thanks for all of your suggestions/improvements.

Abhinav1217:
I have some question.
 Does this work in F25? I can't see much difference on laptop. Even on VM, there is some difference on lower resolution, But then things looked bit weird and out of place. 

2. Using themes are overriding it. on the lower resolution VM, I installed evopop theme, And then it was overrided.

I am currently with satya, This seems personal preference and is not really necessary to be included.

bddenhartog:
> This seems personal preference and is not really necessary to be included.

After some thinking, I too agree with this. This, and any other personal-flavor plugins we have, really belong in the userspace and not in Fedy.

PR #379, opened 2016-06-30
Update install.sh for plugins/arcthemesolid.plugin/ for Fedora 24

jvb187:
Update install.sh for plugins/arcthemesolid.plugin/ for Fedora 24

Repo download path changed for Fedora 24:
http://download.opensuse.org/repositories/home:Horst3180/Fedora_24_standard/home:Horst3180.repo

### DIFF
--- a/plugins/arcthemesolid.plugin/install.sh
+++ b/plugins/arcthemesolid.plugin/install.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-dnf config-manager --add-repo=http://download.opensuse.org/repositories/home:Horst3180/Fedora_$(cat /etc/fedora-release | grep -o "[0-9]*")/home:Horst3180.repo
+dnf config-manager --add-repo=http://download.opensuse.org/repositories/home:Horst3180/Fedora_$(cat /etc/fedora-release | grep -o "[0-9]*")_standard/home:Horst3180.repo
 dnf -y install arc-theme-solid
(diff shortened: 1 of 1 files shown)


plugins/arcthemesolid.plugin/install.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-dnf config-manager --add-repo=http://download.opensuse.org/repositories/home:Horst3180/Fedora_$(cat /etc/fedora-release | grep -o "[0-9]*")/home:Horst3180.repo
+dnf config-manager --add-repo=http://download.opensuse.org/repositories/home:Horst3180/Fedora_$(cat /etc/fedora-release | grep -o "[0-9]*")_standard/home:Horst3180.repo
```
satya164: Does this work on older Fedora versions? If not can we a add a check?

satya164: Does this work on older Fedora versions? If not can we a add a check?

jvb187:
I added a simple if then to check for 24

plugins/arcthemesolid.plugin/install.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
 
-dnf config-manager --add-repo=http://download.opensuse.org/repositories/home:Horst3180/Fedora_$(cat /etc/fedora-release | grep -o "[0-9]*")/home:Horst3180.repo
+if [[ $(cat /etc/fedora-release | grep -o "[0-9]*") = 24 ]]; then
```
satya164: can we do a greater than equals check instead of just equals?

jvb187: Updated to greater than or equals to

PR #381, opened 2016-07-20
Update VirtualBox to 5.1

bugaevc:
As seen in http://download.virtualbox.org/virtualbox/rpm/fedora/24/x86_64/, there now is a VirtualBox 5.1 package.


Abhinav1217:
VirtualBox can be installed via simple dnf install. It is now in RPMFusion repos itself. Fedy has not been providing this plugin for past many releases. The Plugin is there in the repo as an archive.

PR #382, opened 2016-07-20
Add the powerline plugin

bugaevc:
Though powerline's available in the Fedora repos, it requires some configuration to set up. [Add power to your terminal with powerline](https://fedoramagazine.org/add-power-terminal-powerline/) is of some help (esp. the comments), but I still had to figure a lot of stuff out.

This is meant to be the ultimate solution to installing powerline as a bash prompt. I'd be glad if someone familiar with setting it up for use with vim, tmux, etc would come and implement that.

Please do review the scripts as I'm bad at scripting in bash.

### DIFF
--- a/plugins/powerline.plugin/install.sh
+++ b/plugins/powerline.plugin/install.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+dnf list installed powerline || run-as-root dnf -y install powerline
+
+# See comments on https://fedoramagazine.org/add-power-terminal-powerline/
+
+mkdir ~/.config/powerline
+cat > ~/.config/powerline/config.json <<'EOF'
+{
+  "ext": {
+    "shell": {
+      "theme": "default_leftonly"
+        }
+    }
+}
+EOF
+
+grep powerline ~/.bashrc || cat >> ~/.bashrc <<'EOF'
+
+# If powerline is installed and the terminal supports it
+# (e.g. ttys have problems with Powerline fonts), start it
+if command -v powerline-daemon &> /dev/null && [[ $TERM == xterm* ]]; then
+    powerline-daemon -q
+    POWERLINE_BASH_CONTINUATION=1
+    POWERLINE_BASH_SELECT=1
+    . /usr/share/powerline/bash/powerline.sh
+fi
+EOF
--- a/plugins/powerline.plugin/uninstall.sh
+++ b/plugins/powerline.plugin/uninstall.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+rm -rf ~/.config/powerline
+# leave ~/.bashrc as is
+run-as-root dnf -y remove powerline
--- a/plugins/powerline.plugin/metadata.json
+++ b/plugins/powerline.plugin/metadata.json
@@ -1,0 +1,17 @@
+{
+	"icon": "utilities-terminal",
+	"label": "Powerline",
+	"description": "Powerful, flexible and beautiful bash prompt.",
+	"category": "Tweaks",
+	"scripts": {
+		"exec": {
+			"label": "Install",
+			"command": "bash install.sh"
+		},
+		"undo": {
+			"label": "Uninstall",
+			"command": "bash uninstall.sh"
+		},
+		"status": { "command": "bash status.sh" }
+	}
+}
--- a/plugins/powerline.plugin/status.sh
+++ b/plugins/powerline.plugin/status.sh
@@ -1,0 +1,3 @@
+#! /bin/bash
+
+exec test -e ~/.config/powerline


bugaevc:
You would also need a font like Droid Sans Mono Regular for it to look decent.

bugaevc:
> You would also need a font like Droid Sans Mono Regular for it to look decent.

Nah, it actually looks fine with the defaults. I've been over-customizing stuff.

bugaevc:
I'd really like to get this merged (and rolled out).

ping @satya164

By the way, how are [developer plugins](https://github.com/folkswithhats/fedy-plugins-developers) supposed to be used? Should I copy them manually to `/usr/share/fedy/plugins`? I'd prefer to have an option in the Fedy app to fetch additional plugins.

satya164:
cc @bheru27 

@bugaevc you can put in in `~/.local/share/fedy/plugins`

@bddenhartog was working on bringing them to main repo - https://github.com/folkswithhats/fedy/pull/401

bheru27:
What? This sounds nice

bddenhartog:
@bugaevc I'll have time to review this tomorrow

bddenhartog:
I made an issue for this here: #427

bugaevc:
Okay, but I'll actually add one more change -- replace

`if command -v powerline-daemon &> /dev/null && [ $(dirname $(tty)) == /dev/pts ];`

with 

`if command -v powerline-daemon &> /dev/null && [[ $TERM == xterm* ]];`

Abhinav1217:
@bugaevc Any progress on this. Also, what is the purpose of line 7 - 15. I read all the comments and still can't figure out why it is needed. You could simply target `$HOME/.config/powerline`.

bugaevc:
I've made the requested change (rebased my work), now it's up to repo owners to merge this.

Lines 7-15 install powerline configs for every user on system -- `$HOME/.config/powerline` would only be for the one who clicked the button.

satya164:
What if `home/` is not the default home directory? Anyway, I think just setting it up for current user is better, and less invasive, since there's no way to opt-out.

satya164:
Also I think writing the file needs to be done with non-root user, or `chown` it after writing. Otherwise it'll not be editable by the user.

bugaevc:
The other problem with `$HOME` is that the script is being run via `pkexec`, which resets `$HOME`, `$USER` and everything to point to root's data. We can only use `PKEXEC_UID` (= 1000) to find the original user (`awk` over `/etc/passwd`?)

plugins/powerline.plugin/install.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+dnf -y install powerline
+
+# See comments on https://fedoramagazine.org/add-power-terminal-powerline/
+
+for folder in /root /home/*; do
```
bddenhartog: As @satya164 mentioned in the PR discussion, let's refactor this so that we're only setting the current user up for `powerline`. You can get the user ID from `PKEXEC_UID`, and then, to get the username:

`getent passwd "UID_HERE" | cut -d: -f1`

plugins/powerline.plugin/install.sh
```diff
@@ -0,0 +1,41 @@
...
+    # check for /home on a separate partition
+    if [ $folder == /home/lost+found ]; then
+        continue
+    fi
+    # check for no home folders at all
+    if [ $folder == /home/\* ]; then
+        continue
+    fi
+
+    mkdir -p $folder/.config/powerline
```
bddenhartog: Make sure this belongs to the current user, instead of `root`.

plugins/powerline.plugin/install.sh
```diff
@@ -0,0 +1,41 @@
...
+
+    cat > $folder/.config/powerline/config.json <<'EOF'
```
bddenhartog: Make sure this belongs to the current user, instead of `root`.

plugins/powerline.plugin/uninstall.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+for folder in /root /home/*; do
+    rm -rf $folder/.config/powerline
+done
+
+rm -f /etc/profile.d/z-powerline.sh
```
bddenhartog: Fix this path after refactoring `/etc/profile.d/z-powerline.sh` to a user-specific solution.

plugins/powerline.plugin/uninstall.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+for folder in /root /home/*; do
```
bddenhartog: Refactor this to only remove the current user's configuration.

Abhinav1217:
@bugaevc I see your point for all users, and  limitation with $HOME, I don't think Fedora or korora uses any other configuration than `/home` for users, so that should not be an issue.

But I just checked what @satya164  said about non-root user. The file generated by this plugin is not editable by me afterwards. If  `chown` is not possible then we should play with permissions so that users can make changes afterwards.

I will do some more tests and tell you. Maybe I will make one more change request before merging this.

bugaevc:
@satya164 @Abhinav1217 yes, ownership will need to be fixed. Would be doing e.g. `chown -R $(basename $folder) $folder/.config/powerline` (using home folder name to find user's name) be okay?

bddenhartog:
After thinking about this a bit more, I see a problem in a multi-user environment and user-specific configuration:

- `UserA` installs this plugin, which installs the `powerline` package via `dnf`.
- Some time later, `UserB` decides that they want the config too, so they install the plugin (which skips package installation, as it's already installed).
- `UserB` later decides that they do not want to use it anymore, so they remove the plugin. This removes the `powerline` package from the system and now `UserA`'s installation is broken.

Do we care about this? `UserA` will need to reinstall the plugin because of `UserB`'s removal, which isn't ideal. Perhaps this is indicative of a plugin that doesn't belong in Fedy.

Abhinav1217:
@bugaevc Logic seems right, user name can be fetched from folder in home. @bddenhartog has rightly pointed issues with multi-user environment. Uninstall script needs to be rewritten to remove only config file in multiuser environment, not powerline itself. 

You can use http://www.shellcheck.net/ or http://explainshell.com/ to check bash scripts for error. 

Thanks @bddenhartog , Looks like I need to create some multi-user test VMs.

bddenhartog:
> Looks like I need to create some multi-user test VMs.

Yeah, I was thinking that this might be a problem with other plugins too. Don't go to too much trouble - I'm building out automated integration tests that will cover cases like this, and be ran on each plugin on each push.

bddenhartog:
@bugaevc any update on this?

bugaevc:
You see, the way I imagine this (the plugin and Fedy in general) being used is someone setting up the new system after installing it. The person is likely the only user, and if not, at least the one who maintains the system (or is going to do so), in other words, the sysadmin.

You seem to think of the traditional multi-user environment, where all users are equal and each one only changes their own settings. (Yet to use Fedy they would all have access to the root account...)

It would be possible, as you suggest, to only change the settings for the user that clicked/tapped the button, but I think it misses the point.

Furthermore, that way Powerline — even when there's a single user — wouldn't be enabled for the root "user", while it certainly should be.

The precedent here is the "color prompt" plugin  doing a very similar action (setting up a fancier prompt) — and it does set it for all users.

That being said, file permissions should certainly be fixed.

satya164:
>  I don't think Fedora or korora uses any other configuration than /home for users, so that should not be an issue.

I don't think relying on defaults is a good strategy. Plugins have to be flexible to adapt to custom configurations. If not, they aren't high quality enough to be in core.

> The person is likely the only user, and if not, at least the one who maintains the system (or is going to do so), in other words, the sysadmin.

It's likely, but if the user is the single one and has multiple use accounts, it's just a matter of copying the config file.

> Yet to use Fedy they would all have access to the root account

Fedy runs as normal user. You don't need to root to use Fedy. Many plugins are global, so they need root.

> The precedent here is the "color prompt" plugin doing a very similar action (setting up a fancier prompt) — and it does set it for all users.

It certainly does, but it's pretty straightforward. To setup powerline for all users the logic is inaccurate, so might not work for all configurations. Also, it has to change files in home directory for all users and it won't work when I create a new account. It's a pretty different experience from the color prompt plugin.

If you can come up with a way to set it up for all users with their own home directory configuration, I guess it's okay.

bugaevc:
Generally, I agree, but

> Fedy runs as normal user. You don't need to root to use Fedy. Many plugins are global, so they need root.

Can you point out a single plugin that doesn't need root access to install or is not global? Even if you can, the absolute majority are global and need root.

satya164:
> Can you point out a single plugin that doesn't need root access to install or is not global? Even if you can, the absolute majority are global and need root.

The rewrite of Fedy completely was made to properly support single use plugins without root access unlike previous version where it was a pain. Currently non of the plugins do, but that doesn't mean such plugins should not be part of Fedy.

That's not the point here though. The main issue is the method of setting up powerline for all users is not flexible to accomodate different kinds of configurations. To my knowledge, none of other plugins assume a certain configuration and try to work with configuration they have, at least the ones I wrote do. Whereas to set it up a single user, it's straight forward you need to install it with `run-as-root install powerline` and do the rest of the setup as normal user. You have the `$HOME` veriable in environment for current user, so no hacky way to get home directory.

bddenhartog:
@bugaevc Any update on this?

bugaevc:
Yes. I updated the code & rebased it on top of the current master.

It now only installs the configuration files for the current user. Since there's no user-local drop-in directory for environment files (correct me if there is), I have to append the code to `~/.bashrc`, and since there is no easy way to remove the code on uninstallation, I leave it there.

Please let me know if there's a better way.

plugins/powerline.plugin/install.sh
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+}
+EOF
+
+grep powerline ~/.bashrc || cat >> ~/.bashrc <<'EOF'
+
+# If powerline is installed and the terminal supports it
+# (e.g. ttys have problems with Powerline fonts), start it
+if command -v powerline-daemon &> /dev/null && [[ $TERM == xterm* ]]; then
```
Abhinav1217: This should not have any problem when left after uninstallation. But I will comment after few more tests. because if this portion do run when someone uninstalls powerline, whole bashrc will fail.

bddenhartog:
I'm really not sure this plugin belongs in core, simply because the configuration for powerline should be defined by the user. If this configuration were to be in a plugin, I would want proper support for users of all (popular) shells, which would mean supporting the various runcoms. I'm also not a big fan of "uncollected garbage" (the lines left in the runcom after uninstallation).

Thoughts?

Abhinav1217:
I think we can keep this as plugin. This is a very generic configuration that doesn't do much, and doesn't break anything. F25 does preinstall powerline but it does not generate the config file. You have to create it manually, A little more configurations than just "theme": "default_leftonly" would be more meaningful but not a big issue at this time.

bugaevc:
Does F25 really preinstall Powerline? If so, could we get it to be enabled by default (without Fedy), 'cause it's so awesome, and the default plaintext prompt is so dull?

bddenhartog:
> F25 does preinstall powerline but it does not generate the config file. You have to create it manually, A little more configurations than just "theme": "default_leftonly" would be more meaningful but not a big issue at this time.

Okay, but as it stands, as someone who uses ZShell, this pull request does nothing for me - In fact, I would have to know the inner workings of this plugin, and source the script manually. 

So it generates the powerline config file, something a user is almost _sure_ to modify further in the future, saving me from typing out the 40 or so characters. To me, this is a bad idea for two reasons:

- Anyone not using `bash` (a very common occurrence) needs to know how this plugin works and what it does, so they can source the script manually. Such a user is probably not going to be intimidated by the introductory docs / walkthroughs, and would probably consider their `powerline` configuration as a personal file that they would create and maintain the same way they do their `.*shrc` and other dotfiles.

- It creates a basic `powerline` configuration, which new users may be happy with - assuming they are using `bash`. However, this abstracts away the configurability of `powerline` -- new users using this plugin may not even know what powerline really is, or how to further modify it; unlike if they looked up documentation like they would normally have to.

So IMHO this plugin adds very little value, and in fact is detrimental (or at the very least, problematic) for the user. It would be like me including my `dotfiles` as a plugin - I use `zsh` and other specific tools, and my dotfiles are configured for me. It doesn't make sense to include them as a plugin.

Abhinav1217:
@bugaevc  Fedora 25 onwards Powerline will be preinstalled. But config file has to be manually created. I can't find it in the official changelog but I have it on every F25 installation I have done in past few month. I do remember a G+ post about it when the beta was released. I Personally don't like it. It takes up too much space on terminal and it is much tougher to configure than normal $PS1 one. 

@bddenhartog We already have a plugin `Fancy Bash Prompt` How about we modify that to incorporate this if Powerline is found. It will make the message more clear. Only those who want Bash to be fancier will use it. 

Another thought is if we keep it as ` Configure system wide powerline prompt ` tweak, then we can add powerline-tmux, powerline-vim zsh equivalent, and have some basic configuration installed for all.  We can also add condition blocks like **if zsh found install powerline zsh config** to reduce unnecessary stuffs installation.

bddenhartog:
> Another thought is if we keep it as ` Configure system wide powerline prompt ` tweak, then we can add powerline-tmux, powerline-vim zsh equivalent, and have some basic configuration installed for all.  We can also add condition blocks like *if zsh found install powerline zsh config* to reduce unnecessary stuffs installation.  This will make this plugin be more meaningful. 

I guess I'd be okay with this, or with renaming the plugin as "Powerline configuration for Bash". The thing with this, along with plugins doing similar actions, is that they won't exist in `core` after the community repository is opened and integrated, so I have a hard time pushing to merge this in now.

satya164:
Personal opinion: I hate powerline and think it's ugly. I like how our current fancy bash prompt is minimal. I use an even more minimal prompt personally (with zsh).

bheru27:
Well i like powerline but again, you have to tweak it and do some personal stuff on it, which beats the purpose of fedy AND the purpose of powerline and its different usages (zsh,vim...), I don't know man, feels like if we put one for installing oh-my-zsh or antigen(for zsh), you have to tweak it for your needs. Looks cool tho, I would prefer you put it on your github, and link it there so it doesn't disappear in case someone wants to get it

bddenhartog:
> Looks cool tho, I would prefer you put it on your github, and link it there so it doesn't disappear in case someone wants to get it

@bugaevc This is a good idea, and when I finish the major rework to Fedy's internal structure, we'll have a "community repository" -- a centralized location for user-submitted plugins. These will show up in the app like existing plugins, and this would be a great place for installing Powerline and providing a basic preconfigured environment.

Closing this PR for now.

PR #384, opened 2016-08-15
Add GitKraken app 

lobocode:
GitKraken is a sophisticated git client. I added and tested. So far, everything is ok.


delphiki:
It's already present in the [fedy developers plugins repo](https://github.com/folkswithhats/fedy-plugins-developers)

Abhinav1217:
This is available in https://github.com/folkswithhats/fedy-plugins-developers . If your plugin is a better version of the one we already have, then submit PR there. Thanks.

PR #385, opened 2016-08-23
Add support for Moka icon theme

stephenfin:


satya164:
Is Moka actively maintained now a days? Last time I spoke to sam, he told me that it's not.

stephenfin:
> Is Moka actively maintained now a days? Last time I spoke to sam, he told me that it's not.

There are repos published for Fedora 24, so I would assume so?

bddenhartog:
I've reached out to @snwh to get a status update so that a decision can be made about this PR.

bddenhartog:
@stephenfin unfortunately, this is being closed due to the infrequency with which the moka icon repository is maintained. Including projects that are likely to not work with a future version simply causes issues in the future. You can add this plugin locally by adding the `mokaicontheme.plugin` directory to `~/.local/share/fedy`.

PR #388, opened 2016-09-14
Adding a plugin for JetBrains PyCharm Community Edition

andrewderekjackson:
Adding a plugin for JetBrains PyCharm Community Edition

I'm using the following copr repo which, you know, works for me but you might have an opinion about whether that's suitable or not - I defer to your superior wisdom :-)

https://copr.fedorainfracloud.org/coprs/phracek/PyCharm/

Awesome app by the way!


satya164:
This already exists in the developer repo - https://github.com/folkswithhats/fedy-plugins-developers/tree/master/pycharmcommunity.plugin

PR #401, opened 2016-11-09
Restructure plugin categories.

### LINKED ISSUE #397: Better separation of concerns for categories and applications
## Overview

I started using Fedy shortly after moving to Fedora over a year ago. It wasn't until a few weeks ago that I discovered [folkswithhats/fedy-plugins-developers](https://github.com/folkswithhats/fedy-plugins-developers). I'm opening this issue to gain clarity as to why the two are maintained separately, and propose a solution for absolving the need to do so. The immediate issues I see with keeping certain plugins separate from the main app repo are listed below:
- **Inaccessibility**. I can't remember exactly how I stumbled upon it, but the developer plugins repository is not made visible... anywhere, as far as I can tell.
- **Duplication**. This causes unecessary maintenance overhead. There are several plugins which exist in both repositories:
  - Atom
  - Brackets
  - Oracle JDK
  - Sublime text (although the _developers_ repo contains v2, and _this_ repo contains v3)
  - Visual Studio Code
## Proposed solution

I think using additional categories is the right solution here:
- Apps (contains standard applications, like browsers, music players, games, etc)
- Developer Apps (contains IDEs, editors, SDKs, etc - things specific to developers)
- Utilities (contains things like system libraries, e.g. `libdvdcss`, multimedia codecs, fonts, etc)
- Themes (no change in category meaning)
- Tweaks (no change in category meaning)

This will allow for easy integration of the existing `fedy-plugins-developers` plugins, and will help separate some of the overlapping categories with the current layout.

I am, of course, happy to do this myself and submit a PR -- I just wanted to see if this is something that would be beneficial/appreciated before I got started.

bddenhartog:
- Add two new categories: Developers, Utilities
- Incorporate all plugins from fedy-plugins-developers. That repository
  is no longer needed with this patch.
- Reorganize existing plugins into the new categories
- Adjust label on Eclipse for Automotive Software Developers to fix the
  install button getting cut off with the default window size
- Rename 'sublimetext.plugin' to 'sublimetext3.plugin' in order to
  differentiate between Sublime Text version 2 and version 3

Closes #397


bddenhartog:
Additional thoughts:

It looks like the installer currently pulls some plugins that aren't included in this repository by default (Atom, and Visual Studio Code, for example). What is your build process for the RPMs that are installed via [fedy-installer](https://github.com/folkswithhats/folkswithhats.org/blob/master/fedy-installer)? IMHO, it should be updated immediately after this patch so that future installs can take advantage of the newly incorporated plugins.

satya164:
Thank you. Would it be possible to preserve the commit messages for plugins from the developer repo? The way to do it will be,

- Include the developer repo as a submodule
- Un-submodule it - http://stackoverflow.com/questions/1759587/un-submodule-a-git-submodule
- Move files as needed

cc @Abhinav1217

bddenhartog:
Sure, I'll take care of that shortly!

bddenhartog:
Sorry for that temporary close - rebasing back to the original commit (where I had branched off of `master`) caused GitHub to think this branch had already been merged.

In addition to the initial summary provided by this PR, this patch now also brings in the entire history of the developer plugins.

satya164:
I'll leave merging to @Abhinav1217

bddenhartog:
Alright, sounds good. I should also recommend that something is done with the `fedy-plugins-developers` repo; removing it seems unnecessary, but perhaps edit the description and add a README that highlight that it is no longer under development and the plugins are included in Fedy core.

Abhinav1217:
Would it be ok if I merge this after F25 update? I have already bundled it and currently scrutinizing all the scripts and links used within them.  I am also setting up new server to host the repos and website.

`fedy-plugins-developer` was suppose to contain all the plugins and `fedy` repo was suppose to contain app and only copy of those plugins which we would bundling as default with that release. Plugin folder is included in repo for those who would prefer to build fedy from github for testing. I try my best to redirect PR to appropriate repos, but sometime I get lazy and just accept it there itself.   I will edit the README to clear purpose of each repos. Thanks for insight.

bddenhartog:
I guess it just seems strange to me that:

- There's the fedy repo with _some_ of the plugins that are included with the RPM

- There's a separate repo with _some_ of the plugins included in the RPM, and more that aren't, but are probably useful to include in the main build and repo

- The installer installs some predetermined mixture of these two things.

This makes it challenging for Fedy users, who are often probably developers, to find those additional plugins that might be useful to them (such as IDEs). This also makes it difficult for users to reason about the Fedy application - what's installed on their computer doesn't match what they see in the repository.

bddenhartog:
In regards to merging this after you release the new version, I'm totally fine with waiting and working on whatever timeline you think is best; but then I have to ask, is release development done outside of this repository?

satya164:
@bddenhartog I'm not sure how it's being done now, but there's a packaging repo - https://github.com/folkswithhats/fedy-packaging

I think better to keep packaging repo separate from core

bddenhartog:
@satya164 Yes, a separate repository for the RPMs makes sense. What doesn't make sense, is that if you look in the SOURCES directory and uncompress a file there, you'll see that the bundled plugins directory includes plugins from this repo and _some_ from the developer repo.

What I'm saying is that:

- This repo should include _all_ available plugins for fedy, or none (e.g. the changeset provided by this PR)

- The resulting RPM should included _all_ available plugins for fedy, or none.

Currently, it looks like packaging is done separate from both fedy (core) and fedy-plugins-developer. This makes for inconsistencies. Am I making any sense to either of you,  or is this just not something that you see as a problem?

satya164:
> What doesn't make sense, is that if you look in the SOURCES directory and uncompress a file there, you'll see that the bundled plugins directory includes plugins from this repo and some from the developer repo.

It was that way initially, I don't maintain it anymore so I don't have context

> Builds should simply bundle the current "latest/stable" version of this repository without worrying about what plugins are included - it shouldn't care. Your packaging repo definitely shouldn't have its own plugin sources.

Totally agree with you on this

> Build tasks could probably even be moved into this repo, but that's not necessary.

If build tasks are packaging related, I prefer them to be separate

bddenhartog:
> If build tasks are packaging related, I prefer them to be separate

Personally, I don't even see the need for manually building the RPM. [The build script](https://github.com/folkswithhats/fedy-packaging/blob/master/build) is nothing special and could easily be set up on a server for automated deployment and building - first downloading this repository, then performing the various actions needed based on metadata. This could be triggered via a webhook on each commit to `master`, so that the entire workflow is automated. Of course, I'm not familiar with uploading a package to RPMFusion, so perhaps there is still some manual work needed, but I think we could make a few small changes and greatly reduce the workload behind releasing fedy.

bddenhartog:
@satya164 @Abhinav1217 I think I can actually fix the historical paths so that the imported history (from `fedy-plugins-developer`) references the new plugin locations. I'll work on this and have a patch out by the weekend.

**I will be force pushing this branch**, so if you have pulled it locally and made changes, please rebase.

satya164:
@bddenhartog are you still willing to work on this? I can give you push access to the repo if you want.

bddenhartog:
Yes, absolutely. I'm out of town until tonight but can finish this up then.

bheru27:
What in the he- 
nvm, remember i only git pull from the repo to build it, you tell me when its done

bugaevc:
What's going on?

bddenhartog:
@bugaevc I deleted that branch for three reasons:

- It integrated the `fedy-plugins-developers` history without reassigning file paths, which -- while it isn't a problem -- definitely isn't ideal in terms of tracking history.
- It was from my own fork, made before I became a collaborator on the project.

Both repositories had seen significant changes, so instead of reverting the feature branch (from my fork), updating, redoing everything the proper way and force-pushing, I decided to simply delete it and re-implement it from a feature branch on this repository.

bddenhartog:
@bugaevc The new PR is here: #416

bugaevc:
@bddenhartog awesome, thanks!

PR #413, opened 2016-12-27
Add: Popcorn-Time.sh

cadesalaberry:
Any comment is welcome!

Cheers

### DIFF
--- a/plugins/popcorntime.plugin/install.sh
+++ b/plugins/popcorntime.plugin/install.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/popcorntime"
+
+if [[ "$(uname -m)" = "x86_64" ]]; then
+  ARCH="64"
+else
+  ARCH="32"
+fi
+
+mkdir -p "$CACHEDIR"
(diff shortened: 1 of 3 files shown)


bddenhartog:
@Abhinav1217, doesn't this PR leave the downloaded file on the user's filesystem? Just gave it a quick scan but perhaps this was merged in hastily.

satya164:
@bddenhartog All plugins do that. This allows to carry them over to a new machine if needed. Probably a new plugin to clear Fedy's cache will be useful.

bddenhartog:
@satya164 This particular installation script isn't using any kind of cache directory, though, it's simply downloading to the user's current directory. If I'm in `/var/log/whatever`, it downloads and leaves the compressed file there. If I'm in `/home/user/mydir`, it downloads and leaves it there.

satya164:
it cd's to cache directory before the download - https://github.com/folkswithhats/fedy/pull/413/files#diff-3602ef5bf528974e5f9931d8b69876a9R12

bddenhartog:
Wow, I have no idea how I missed that. Must need to change my contacts. Whoops!

satya164:
xD

PR #415, opened 2016-12-29
Spotify crashes after logging in

### LINKED ISSUE #412: spotify crashes
I am running a fresh install of Fedora 25. Starts up fine. When I type my username/password, it crashes upon submission.

```bash
[robert@robert-desktop applications]$ spotify 
spotify: /lib64/libssl.so.1.0.0: no version information available (required by spotify)
spotify: /lib64/libcrypto.so.1.0.0: no version information available (required by spotify)
spotify: /lib64/libcurl.so.4: no version information available (required by spotify)
/usr/lib64/spotify-client/spotify: /lib64/libssl.so.1.0.0: no version information available (required by /usr/lib64/spotify-client/spotify)
/usr/lib64/spotify-client/spotify: /lib64/libcrypto.so.1.0.0: no version information available (required by /usr/lib64/spotify-client/spotify)
/usr/lib64/spotify-client/spotify: /lib64/libcurl.so.4: no version information available (required by /usr/lib64/spotify-client/spotify)
[1223/015123:ERROR:main_delegate.cc(752)] Could not load cef_extensions.pak
/proc/self/exe: /lib64/libssl.so.1.0.0: no version information available (required by /proc/self/exe)
/proc/self/exe: /lib64/libcrypto.so.1.0.0: no version information available (required by /proc/self/exe)
/proc/self/exe: /lib64/libcurl.so.4: no version information available (required by /proc/self/exe)
[1223/015123:ERROR:main_delegate.cc(752)] Could not load cef_extensions.pak
[1223/015147:ERROR:PlatformKeyboardEvent.cpp(94)] Not implemented reached in static bool blink::PlatformKeyboardEvent::currentCapsLockState()
Segmentation fault (core dumped)
```

stanfieldr:
This seems to be an issue [here](https://github.com/negativo17/spotify-client/issues/11) and solved [here](https://github.com/negativo17/spotify-client/issues/11#issuecomment-268639736) which also fixes #412. I also added cef_extensions.pak which was showing as an error before when booting from terminal.


Abhinav1217:
While using deb is convenient workaround, only way to update app would be to uninstall it and reinstall it. I did not have issue in F25 using xorg session, only in wayland. (maybe because they removed status icon apis).  Are you sure there are no copr repos to provide updates? You can host a copr yourself too.

stanfieldr:
Yea you're probably right, wouldn't be easy to update.

Abhinav1217:
I just found out that spotify is in official fedora rpmfusion repos. How about we have a script which installs from official repos in F25 and uses original installer in F24 and below.

PR #416, opened 2017-01-12
Integrate developer plugins, restructure plugin categories

bddenhartog:
This patch:

- Integrates the plugins and history from the `fedy-plugins-developers` repository, which can (upon PR acceptance) be empty/deprecated and users redirected to fedy core. The history from the integrated repository has been modified to alter paths to the plugin files, moving them into the `plugins/` directory to match the structure of the core project (this repository). This does not create a _perfect_ history, unfortunately, as it appears the `fedy-plugins-developers` repository history was originally yanked/forked from core, with several files removed (but not as part of a commit). I'm not sure if this is something we really care about.

- Adds two new plugin categories: `Editors and IDEs` and `Utilities`. The latter is intended to hold plugins like system utilities or development libraries, such as AV codecs or SDKs. The eventual goal here is to keep the `Apps` categories clear of things that aren't GUI applications, and to add better category lists so it is easier for users to find what they want to install.

- Refactors the Eclipse Automotive plugin to shorten the name displayed in the Fedy application. Previously, this name extended beyond the default width for the Fedy application, causing a poorer user experience.

- Refactors the icon names for the Sublime Text plugins to explicitly specify the versions (2 & 3)

**IMPORTANT**
Upon merging, please add `Closes #397` to the bottom of the commit message, so that the related issue is closed. It is important that a **non-ff** merge is performed. **Do not** perform a squash merge. Since the Github UI prevents a non-ff merge (because it is not enabled for this repository), this merge should be performed from the command line:

```text
# Go the the repository on your system
cd /path/to/the/fedy/repo

# Update remote refs and the master branch
git checkout master && git pull

# Merge this feature branch
# Make sure you add "Closes #397" under the automatically-generated message!
git merge --no-ff feature/397/integrate-developer-plugins

# Push up to the remote
git push
```

Relates to #397

PR #418, opened 2017-01-16
Recategorize development tools

danstewart:


bddenhartog:
I'm okay with this, but the initial goal with the final category name was to allow us to more accurately categorize our plugins, with the future goal of reworking the UI to account for any number of tabs (or better yet, refactor the tabs into a dropdown-esque or searchable field).

This would allow us to extend the included plugins to a much larger quantity, while still keeping a sane UI that can be easily navigated.

danstewart:
@bddenhartog All of them apart from 2 (GitKraken & Arduino) are editors or IDEs.
Might make more sense to split them up again when there are more plugins 😄

bddenhartog:
Fair point 👍

danstewart:
That being said - I just noticed SmartGit and Gnome Development Tools are both under Utilities - maybe GitKraken and Arduino would be better there? The lines between categories are a little fuzzy.
I think it would be preferrable to have all development related stuff away from non development stuff.

bddenhartog:
> The lines between categories are a little fuzzy.

Agreed. I intend to revisit every plugin and provide saner categorization in the future. "Utilities" was provided by #416, and you're right, GitKraken should have been included in there initially. "Arduino" is actually [an IDE](https://www.arduino.cc/en/Main/Software), so it belongs under "Development Tools" in this PR.

bddenhartog:
@danstewart There's more (slightly outdated) info on the initial issue I logged for integrating the dev plugins: #397

danstewart:
@bddenhartog Done.

danstewart:
@bddenhartog Done - tried squashing my commits and somehow missed one but rolled back to your commit renaming visual studio code - no idea how I managed that.

Should Gnome Dev Tools stay where it is?

bddenhartog:
Whoops - I was adding instructions for you to fix your branch that had diverged from `folkswithhats/fedy/master` due to your change on VSC, but I see that you're fixing it another way :)

PR #423, opened 2017-01-19
Fix to Gitkraken plugin file, to show shortcut on menu

WolfangAukang:
The Gitkraken's folder name goes in lowercase. This seems to solve the issue of not showing Gitkraken's shortcut on the menu.


bddenhartog:
Please create an issue describing the need for this, and reference that issue in your commit message (on a second line)

Issue #422, opened 2017-01-18
Enable community plugins


sudoforge:
We receive a lot of issues and pull requests for plugins that we do not, for one reason or another, accept as an official plugin. Usually, this is because the plugin fits a niche role, or there are already alternatives in Fedy, or the plugin isn't actively maintained (or at least, not maintained as actively as we want). Ultimately, this is detrimental to the community as a whole, as users miss out on the discovery and ease of installation of various applications and tools that they might not necessarily be exposed to; additionally, two users who want the same niche application added must each individually add and manage their own Fedy plugin.

In order to ease these issues, we are planning on implementing "community plugins". The implementation details have yet to be planned out, but the initial thought is that you will be able to easily browse and enable community-built plugins via the GUI.

Initial thoughts:

- Both `main` and `community` plugin repositories will be abstracted away from the `fedy` repository. `main` and `community` repositories will be separate from each other, with different standards for maintenance and acceptance. The FWH team will not officially support (add, fix, or otherwise modify) community plugins, but we will review them before acceptance (for security and other reasons).

- The `fedy` repository will include only the UI elements, and the Fedy application will no longer ship with any plugin instructions. Instead, it will call out to the `main` and `community` repositories for the _current_ plugin list, and display as needed. This will allow for dynamic updating of plugin lists, much like users are used to from other package managers.

- The `community` repository will be more lenient in the plugins that are accepted, which means that effectively, if someone wants their niche program or library or some alternative to a `main` plugin to be available via Fedy, they will be able to create a `community` plugin and submit it for review. We'll make sure it's formatted correctly and will work, and as long as everything is on the up-and-up, it will be accepted. We'll also deprecate plugins that become unmaintained, and remove any that need removal (either by the author's request or because it's broken in some sense). We (as a team) will not maintain these plugins, although it is possible that team members maintain their own community plugins (we do use Fedy, after all).

- The `main` repository will continue to be more restrictive in the plugins we accept. Generally, we want to keep general-purpose, popular, useful plugins in `main`. These plugins will be officially supported, meaning that the Fedy team will keep them up-to-date and work on issues as needed. There may be some reorganization of the plugins currently in core when the community repository is available.

**Please feel free to discuss.** Fedy is a community-driven application, and all feedback is appreciated.

danstewart:
Additional comments from @bddenhartog 

[...]

PR #425, opened 2017-01-20
Add JetBrains ToolBox App plugin

delphiki:
This plugin adds the [JetBrains ToolBox App](https://www.jetbrains.com/toolbox/app/), which makes it a lot easier to install/update all the JetBrains apps (PHPStorm, ItelliJ, WebStorm, etc.).
So I don't know if we should keep thoses apps as plugins.

### DIFF
--- a/plugins/jetbrains-toolbox.plugin/jetbrains-toolbox.svg
+++ b/plugins/jetbrains-toolbox.plugin/jetbrains-toolbox.svg
@@ -1,0 +1,21 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns" width="47" height="53" version="1.1"><rect id="backgroundrect" width="100%" height="100%" x="0" y="0" fill="none" stroke="none" class=""/>
+    <title>Slice 1</title>
+    <description>Created with Sketch (http://www.bohemiancoding.com/sketch)</description>
+    <defs>
+        <linearGradient x1="0.633426184%" y1="72.4273392%" x2="100.157409%" y2="27.2015838%" id="linearGradient-1">
+            <stop stop-color="#FF8618" offset="4.3%"/>
+            <stop stop-color="#FF246E" offset="38.17%"/>
+            <stop stop-color="#AF1DF5" offset="98.92%"/>
+        </linearGradient>
+    </defs>
+    
+<g class="currentLayer"><title>Layer 1</title><g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage" class="" id="svg_1">
+        <g id="svg_3" sketch:type="MSLayerGroup">
+            <path d="M40.45402669914946,36.88730387505994 L28.93705322223853,43.193594335594206 L28.93705322223853,38.18452451969123 L40.45402669914946,31.878234059156952 L40.45402669914946,36.88730387505994 L40.45402669914946,36.88730387505994 z" id="Shape" fill="#FFFFFF" sketch:type="MSShapeGroup"/>
+            <g id="Group" sketch:type="MSShapeGroup">
+                <path d="M23.599277843670226,52.71082698580989 L47.19855568734045,39.50743270183997 L47.19855568734045,13.139175440176297 L23.599277843670226,26.34256972414621 L23.599277843670226,52.71082698580989 z" id="Shape" fill="#000001"/>
+                <path d="M28.529489092347855,40.457871589985665 L28.529489092347855,44.41375236767313 L38.82377017958673,38.659743963764065 L38.82377017958673,34.7038631860766 L28.529489092347855,40.457871589985665 z" id="Shape" fill="#FFFFFF"/>
+                <path d="M23.599277843670226,1.9630179135031543e-16 L0,13.139175440176297 L0,39.50743270183997 L23.599277843670226,52.71082698580989 L23.599277843670226,26.34256972414621 L47.19855568734045,13.139175440176297 L23.599277843670226,1.9630179135031543e-16 z" id="Shape" fill="url(#linearGradient-1)" transform=""/>
+            </g>
+        </g>
+    </g></g></svg>
--- a/plugins/jetbrains-toolbox.plugin/jetbrains-toolbox.svg
+++ b/plugins/jetbrains-toolbox.plugin/jetbrains-toolbox.svg
@@ -9,7 +9,9 @@
         </linearGradient>
     </defs>
     
-<g class="currentLayer"><title>Layer 1</title><g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage" class="" id="svg_1">
+    <g class="currentLayer">
+        <title>Layer 1</title>
+        <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage" class="" id="svg_1">
         <g id="svg_3" sketch:type="MSLayerGroup">
             <path d="M40.45402669914946,36.88730387505994 L28.93705322223853,43.193594335594206 L28.93705322223853,38.18452451969123 L40.45402669914946,31.878234059156952 L40.45402669914946,36.88730387505994 L40.45402669914946,36.88730387505994 z" id="Shape" fill="#FFFFFF" sketch:type="MSShapeGroup"/>
             <g id="Group" sketch:type="MSShapeGroup">
@@ -18,4 +20,6 @@
                 <path d="M23.599277843670226,1.9630179135031543e-16 L0,13.139175440176297 L0,39.50743270183997 L23.599277843670226,52.71082698580989 L23.599277843670226,26.34256972414621 L47.19855568734045,13.139175440176297 L23.599277843670226,1.9630179135031543e-16 z" id="Shape" fill="url(#linearGradient-1)" transform=""/>
             </g>
         </g>
-    </g></g></svg>
+        </g>
+    </g>
+</svg>
--- a/plugins/jetbrains-toolbox.plugin/install.sh
+++ b/plugins/jetbrains-toolbox.plugin/install.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+dnf -y install java-1.8.0-openjdk java-1.8.0-openjdk-devel compat-libstdc++-296.i686 compat-libstdc++-33.i686 compat-libstdc++-33.x86_64 glibc.i686 ncurses-libs.i686
+
+CACHEDIR="/var/cache/fedy/jetbrains-toolbox";
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+URL=$(wget "https://data.services.jetbrains.com/products/releases?code=TBA&latest=true&type=release" -O -| grep -o "https://download.jetbrains.com/toolbox/jetbrains-toolbox-[0-9.]*.tar.gz" | head -n 1)
+FILE=${URL##*/}
+
+wget -c "$URL" -O "$FILE"
+
+if [[ ! -f "$FILE" ]]; then
+	exit 1
+fi
+
+tar -xzf "$FILE" -C "/opt/"
+
+mv /opt/jetbrains-toolbox* "/opt/jetbrains-toolbox"
+ln -sf "/opt/jetbrains-toolbox/jetbrains-toolbox" "/usr/bin/jetbrains-toolbox"
+
+cp "$(dirname $0)/jetbrains-toolbox.svg" "/usr/share/icons/hicolor/scalable/apps/"
+gtk-update-icon-cache -f -t /usr/share/icons/hicolor
+
+cat <<EOF | tee /usr/share/applications/jetbrains-toolbox.desktop
+[Desktop Entry]
+Name=JetBrains Toolbox App
+Type=Application
+Exec=jetbrains-toolbox
+Icon=jetbrains-toolbox
+Comment=A control panel for your tools and projects
+Categories=Development;IDE;
+Keywords=Idea;PHPStorm;Web;Toolbox;Jetbrains;IDE;
+StartupNotify=true
+Terminal=false
+StartupWMClass=jetbrains-toolbox
+EOF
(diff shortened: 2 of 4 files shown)


plugins/jetbrains-toolbox.plugin/jetbrains-toolbox.svg
```diff
@@ -0,0 +1,21 @@
...
+            <g id="Group" sketch:type="MSShapeGroup">
+                <path d="M23.599277843670226,52.71082698580989 L47.19855568734045,39.50743270183997 L47.19855568734045,13.139175440176297 L23.599277843670226,26.34256972414621 L23.599277843670226,52.71082698580989 z" id="Shape" fill="#000001"/>
+                <path d="M28.529489092347855,40.457871589985665 L28.529489092347855,44.41375236767313 L38.82377017958673,38.659743963764065 L38.82377017958673,34.7038631860766 L28.529489092347855,40.457871589985665 z" id="Shape" fill="#FFFFFF"/>
+                <path d="M23.599277843670226,1.9630179135031543e-16 L0,13.139175440176297 L0,39.50743270183997 L23.599277843670226,52.71082698580989 L23.599277843670226,26.34256972414621 L47.19855568734045,13.139175440176297 L23.599277843670226,1.9630179135031543e-16 z" id="Shape" fill="url(#linearGradient-1)" transform=""/>
+            </g>
+        </g>
+    </g></g></svg>
```
bddenhartog: Need a `newline` here.

Abhinav1217:
In the description, maybe we could change it to say Toolbox to install and manage jetbrains application. So that users may know they can use that, if we remove all the individual plugins.

delphiki:
Sorry for the delay, I'll update the PR with your feedback asap!

satya164:
> Please create an issue describing the need for this and the reasoning behind removing the existing plugins - it helps when investigating in the future.

Why do we need a separate issue for the PR? For investigation, can't we just look at the PR's description?

> you should add it on a new line, so it isn't part of the git log --oneline output.

Why the restriction? Anyway, we can format the commit message as we want when merging, so this shouldn't be an issue.

bddenhartog:
> Why the restriction? Anyway, we can format the commit message as we want when merging, so this shouldn't be an issue.

It's important to maintain correlation between a patch and an issue. I don't know how you browse repository history, but if I'm looking at a commit from 4 months ago that brought in a bug or malicious code, I need to be able to understand why that commit was made in the first place. 

Commits should **always** contain the issue number they relate to. As for closing issues, yes, that can be added as part of the merge commit by whoever merges the changes in. Squashing is less ideal as git loses all reference to the original author if the squashed commit message is modified. Github continues to track it, sure, but it would be a poor decision to rely on a feature provided by a private third party. 

> Why do we need a separate issue for the PR? For investigation, can't we just look at the PR's description?

Issues are the "central area" around a feature/bug/etc. Pull requests _relate to issues_. People _create issues_ prior to doing work. It is convention to create an issue when there is a need or desire for work, and then create a pull request with the changeset. I see no reason why Fedy shouldn't follow convention.

satya164:
I'll quickly summarize why separate issues and PRs are not better than a single PR

1. The motivation and the details about the PR stay in the issue, not the commit message. You can't read it from the commit message if you want and always have to come to GitHub to read it. Also many times I read commit messages in GitHub UI and it's super inconvenient to open another page just to see what it is.
2. It's better to create an issue when people need to discuss things prior to the PR, so the intent is clear when someone opens a PR. The PR still should contain brief description of the motivation and changes. But if someone has already created a PR, there's no value in creating a separate issue.
3. When merging a commit github automatically adds the PR number in the commit title, or you could add it yourself when merging. So there is no need to require contributors to do it manually.
4. I've never seen any OSS project to strictly require 1:1 correlation between issues and PRs, I've actually seen people turn issues into PRs once the discussion is settled. This has actually kept repos manageable without discussions being split across the PR and the issue.

IMO we shouldn't increase friction for contributors for things which don't add much value. Many of these would have been valid few months ago, but now GitHub has better tools for merging. So we can actually make it easier for contributors and not require any manual work from them.

bddenhartog:
Your entire investigation process relies on:

- Contributors squashing commits when accepting a PR

- Github adding the PR number to the squashed merge commit, or the contributor doing that manually.

... and fails to correlate commits to any kind of informative view outside of that scenario. I can tell you that personally, I merge most often via the command line, and in order to preserve history and original author, I most often perform non-ff merges (merges that bring in the branch's commits and creates a merge commit).

So now this unspoken strategy of relying on github's ui feature and squashing commits (which really loses history - imagine if we did that when I merged history from the developer repo). Additionally, it becomes more difficult to track individual or minute changes because every commit is always squashed into an entirely new commit. Now the refs from the user's remote don't exist in upstream, and their author information is lost (unless the squash is performed via the github ui). IMO this is a very bad strategy to rely upon.

delphiki:
a1549b8621e83854cda79881ca75a953627840db5ab3cf8c2bd2caa29bac5e7c

satya164:
You're overthinking it

bddenhartog:
Again, I really don't think so... I spend my days working on large projects where preserving history matters, and because a strategy like this wasn't always in place, we run into issues from time to time when we need to investigate changes. I also don't use the Github UI to browse history, so something like your commit in #335 that doesn't reference any issue loses all context with viewing in `git log`. PRs that I review will always require issues, and properly formatted commit messages. The only downside here is that users who contribute patches to Fedy must adopt healthy git habits, which... isn't really a downside.

satya164:
I work in large projects too.

I'll give you a simple example, let's say this is the commit you merged,

`[hash] fix some issue in some place (#34)`

Now, how does it even matter if `#34` is an issue or a PR. You go to `#34` and you read the discussion. Why have the discussion split between `#33` and `#34`?

Let's say you didn't squash the commits because it had multiple commits and you want to keep all the history,

`[hash] some initial attempt to fix something`
`[hash] fix some issue in some place (#34)`

Now again, how does it matter if `#34` is an issue or a PR?

At least if issue was same was PR, you get the option to actually keep the issue text in the commit message too.

> like your commit here that doesn't reference any issue

It's a terrible commit because I didn't put enough details there as well as a is of the PR. It wouldn't have made any difference if there was a different issue if I didn't link to it.


> PRs that I review will always require issues, and properly formatted commit messages. The only downside here is that users who contribute patches to Fedy must adopt healthy git habits

Sure, properly formatted commit messages are important. But separate issues for every PR is unreasonable. By that logic, for every commit you push to master, you need to first create a PR for it.

satya164:
And let's be practical, in an open source project, when you have an issue and a PR on the same topic, people tend to comment on both. So your issue isn't the place for discussion as you are saying. Most of the time the PR becomes the place of discussion.

bddenhartog:
> I'll give you a simple example, let's say this is the commit you merged,
> 
> [hash] fix some issue in some place (#34)
> 
> Now, how does it even matter if #34 is an issue or a PR. You go to #34 and you read the discussion. Why have the discussion split between #33 and #34?

I'll ignore the point I made earlier about issues being grounds for discussion around the _why_ and pull requests being grounds for discussion around the implementation for the sake of this example. It doesn't matter if `#34` is an issue or PR in this case - but in order for us to not care about this, issues _always_ need to be converted into pull requests. In order to do that, every committer must use `hub` (as I mentioned earlier, there is no way that I can see to do this in the github ui). You've now brought in an additional software requirement that might not be part of everyone's toolset.

> Let's say you didn't squash the commits because it had multiple commits and you want to keep all the history,
> 
> [hash] some initial attempt to fix something
> [hash] fix some issue in some place (#34)
> 
> Now again, how does it matter if #34 is an issue or a PR?0

The issue here isn't _what_ `#34` references, it's that the first commit might have been made three days before the second one, which actually references the issue/pr. Now when I'm viewing `git log` and see `some initial attempt to fix something`, I have absolutely no way to gain context as to why in the world that commit was made.

> every commit you push to master

Nobody should be pushing to `master` directly. That's actually disabled in the repo settings. Maintainers can't even do it. There's never a valid reason that a change shouldn't be reviewed. Every change that goes into the master branch should be reviewed.

satya164:
> issues always need to be converted into pull requests (otherwise they will have separate IDs). In order to do that, every committer must use hub (as I mentioned earlier, there is no way that I can see to do this in the github ui). You've now brought in an additional software requirement that might not be part of everyone's toolset.

I'm just saying it's unnecessary to create a new issue after a PR is already submitted. Since we agree that it doesn't matter if `#34` is a PR or an issue, if there's already a separate issue for a PR, we don't need to convert it to a PR.

The only thing that matters is the commit message has the number of the issue or the PR `#34`  to follow the discussion when needed.

> The issue here isn't what #34 references, it's that the first commit might have been made three days before the second one, which actually references the issue/pr.

The problem is that there was no issue created before the PR was submitted. Now asking to create an issue after the PR is submitted won't solve the problem, because the commit messages won't reference the issue. Only a rebase or a new commit can add the issue number, it as well be the pr number.

Consider this PR for example. It was created 7 days ago. And now there's an issue regarding this. But you won't be able to follow the issue because there is no reference.

> Nobody should be pushing to master directly. That's actually disabled in the repo settings. Maintainers can't even do it. There's never a valid reason that a change shouldn't be reviewed. Every change that goes into the master branch should be reviewed.

Yeah, certainly not for projects with more than 1 collaborators. But what I meant to say that, you shouldn't have to create an issue if you're doing something small as fixing a typo.

bddenhartog:
> I'm just saying it's unnecessary to create a new issue after a PR is already submitted

I can agree with this, but only because _the issue should have been created before the PR was submitted in the first place_. The reason I'm asking people who have opened these "detached" pull requests to create an issue and update their commits to reference that issue is so the bad behavior can be fixed.

> Since we agree that it doesn't matter if #34 is a PR or an issue

We don't agree that it doesn't matter. I've already explicitly stated that I think converting issues to pull requests, or discussion around the _why_ happening on a pull request is bad. My earlier comment saying that it didn't matter was only made for the sake of the example you were giving.

> The problem is that there was no issue created before the PR was submitted. Now asking to create an issue after the PR is submitted won't solve the problem, because the commit messages won't reference the issue. Only a rebase or a new commit can add the issue number, it as well be the pr number.

Yes, that's exactly what I'm having people do. If a "detached" pull request exists, as part of the review process, I ask the contributor to fix their commit messages. I've provided examples to other people, but @delphiki obviously knew how to do this on his own as he went through and made the requested changes appropriately. (Ace job 👍 btw)

> Consider this PR for example. It was created 7 days ago. And now there's an issue regarding this. But you won't be able to follow the issue because there is no reference.

As mentioned above, there actually is. You can even see them on the issue activity: #428

> But what I meant to say that, you shouldn't have to create an issue if you're doing something small as fixing a typo.

Sure, I can agree with that. I'd consider that a `chore` or `cleanup`, for which creating an issue is probably unnecessary, as all of the information is easily contained within the commit. But we're not talking about a teeny change here, we're talking about plugin additions and removals, other features, etc.

satya164:
> I've already explicitly stated that I think converting issues to pull requests, or discussion around the why happening on a pull request is bad

I never suggested we start doing that. I merely mentioned somewhere that I've seen people do that.

bddenhartog:
I think we agree on the preferred workflow:

- A problem is found or feature desired, so a user creates an issue
- Someone (maybe that user) branches off and starts committing, referencing that issue in each commit
- The committer then opens a pull request, and references the issue in the description of the PR (in order to surface the issue->PR connection in Github)
- That pr is merged or closed

With this, everything is visible on the issue's page.

I don't really think there's any room for argument against this workflow for any repo with more than one contributor. I follow this on my personal, private projects anyway, too, because it's the proper way to go about things.

Once I get contributing guidelines up, I don't think we'll have any problems with people adopting to the workflow.

satya164:
Sounds good. Create an issue for contributing guidelines.

bddenhartog:
> Sounds good. Create an issue for contributing guidelines.

Of course, prior to working on it I'll have an issue up :)

Abhinav1217:
Other discussion aside, @delphiki Did you get time to update the PR with better description.  Do not remove the other jetbrain plugins in your commit as I would prefer to keep them in the github for a while. Call it legacy support or just maintaining history.

@bddenhartog  Please do not suggest plugin removal to the contributors. It is what maintainers and admins are for. Also due to issues in past, it is better to keep legacy plugins in the repo for at least a few months for those who need it, when we stop bundling them in rpms.

delphiki:
@Abhinav1217 I did update the description: https://github.com/folkswithhats/fedy/pull/425/commits/a131bb57d719a4a17cf2b7061f83c797c5a5d456
I can delete the commit removing the jetbrains apps & the issue opened about that if want to.

Adding a deprecation notice for plugins within fedy could be nice, maybe a [Deprecated] tag in the plugins' descriptions?

bddenhartog:
> I can delete the commit removing the jetbrains apps & the issue opened about that if want to.
> 
> Adding a deprecation notice for plugins within fedy could be nice, maybe a [Deprecated] tag in the plugins' descriptions?

I'm against both of these things.

- The Toolbox replaces the other applications. It enables users to manage the installations appropriately; there is no need to retain the plugins.

- Deprecations should be surfaced some other way, and the user should be prompted to remove deprecated plugins from their system. I'll work this into #422, whose description I'm going to be updating today as the scope has been better defined.

Abhinav1217:
@delphiki Please do not remove the older jetbrain plugins from the repo. Just add your new plugin. 

@bddenhartog Keeping old plugins in github does not mean we have to keep pushing them via dnf. We faced same issue with eclipses last year. Therefore from past experience, I want to keep them in Github, even though we can stop pushing them via dnf releases.

bddenhartog:
> Keeping old plugins in github does not mean we have to keep pushing them via dnf. We faced same issue with eclipses last year. Therefore from past experience, I want to keep them in Github, even though we can stop pushing them via dnf releases.

As I've expressed elsewhere on various occasions, I think separating what is installed and what is in the repo is confusing to users. But I digress, this won't be an issue in the future when instruction sets are delivered by the api.

bddenhartog:
> I want to keep them in Github, even though we can stop pushing them via dnf releases.

Also regarding this, what's the worry here? The toolbox app will allow the user to manage their installations.

delphiki:
There you go, I deleted the commit removing the jetbrains apps.

PR #335, opened 2017-01-26
Search option not including name string

satya164:
Hi,

It seems the search option only queries the description.
If you for instance Search for "Flash" in that app tab nothing matches.

Gr, J


satya164:
@Abhinav1217 can you check that this fixes the issue?

Abhinav1217:
No, With your code I am getting this error  

```
(gjs:11875): Gjs-WARNING **: JS ERROR: TypeError: items[0].get_label is not a function
Application<._renderPlugins/</filter@./app.js:599
Application<._renderPlugins/<@./app.js:610
@./app.js:711
```
Then I tries changing value of 0 to 1, It gave same issue.

Using value 2, does not generate errors but it doesn't solve the problem either. Result is same as using original value of 3, i.e. Search only looks in the App descriptions, not in the app Title.

But Using value of 4 solves the problem, ` title = items[4].get_label(),` I will make changes close your pr without committing. Thanks.

Abhinav1217:
It looks like github has added some odd security restrictions. @satya164 I am deleting your branch as I created another branch because I can not add anything directly on master branch. Now there is duplicate branch. It's getting confusing..

satya164:
@Abhinav1217 seems someone changed repo settings?

PR #438, opened 2017-03-05
Switch to official vscode repository

Artilex:
With the addition of [official repositories](https://code.visualstudio.com/updates/v1_10#_official-signed-repositories-for-linux) for Visual Studio Code the use of Mosquito's repository is no longer necessary.

The changes copies the instructions provided at the [vscode website](https://code.visualstudio.com/docs/setup/linux#_rhel-fedora-and-centos-based-distributions).


plugins/visualstudiocode.plugin/install.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 
-dnf copr -y enable mosquito/vscode 
-dnf -y install vscode
+rpm --import https://packages.microsoft.com/keys/microsoft.asc
+sh -c 'echo -e "[code]\nname=Visual Studio Code\nbaseurl=https://packages.microsoft.com/yumrepos/vscode\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" > /etc/yum.repos.d/vscode.repo'
```
satya164: Why the extra `sh -c`? It's already running in bash.

Also can you break this to separate lines instead of a long line? Something like https://github.com/folkswithhats/fedy/blob/master/plugins/touchpad.plugin/setup.sh#L11

plugins/visualstudiocode.plugin/install.sh
```diff
@@ -1,4 +1,6 @@
...
+dnf check-update
```
satya164: Is this necessary?

plugins/visualstudiocode.plugin/uninstall.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-dnf copr -y disable  mosquito/vscode
-dnf -y --setopt clean_requirements_on_remove=1 erase  vscode -y
+dnf config-manager --set-disabled code
```
satya164: We should just delete the repo file like we do in other plugins where we add the repo file manually.

plugins/visualstudiocode.plugin/install.sh
Artilex: No reason really, I hastily copied it over before leaving today. I'll push a more elegant solution. @satya164

plugins/visualstudiocode.plugin/install.sh
```diff
@@ -1,4 +1,18 @@
 #!/bin/bash
 
-dnf copr -y enable mosquito/vscode 
-dnf -y install vscode
+gpgkey=$"https://packages.microsoft.com/keys/microsoft.asc"
+
+rpm --import $gpgkey
+
+cat <<EOF > /etc/yum.repos.d/vscode.repo
+[vscode]
```
bddenhartog: This should be the official `code` instead of `vscode`.

plugins/visualstudiocode.plugin/install.sh
```diff
@@ -1,4 +1,18 @@
...
+name=Visual Studio Code
+baseurl=https://packages.microsoft.com/yumrepos/vscode
+enabled=1
+gpgcheck=1
+gpgkey=$gpgkey
+EOF
+
+dnf -y install code
+
+sed -i 's/Icon=code/Icon=vscode/g' /usr/share/applications/code.desktop
```
bddenhartog: The supplied icons seem to be named `code` instead of `vscode`, so I don't think this would properly display the icon included with the official repository.

plugins/visualstudiocode.plugin/metadata.json
```diff
@@ -14,6 +14,6 @@
 			"label": "Remove",
 			"command": "run-as-root -s uninstall.sh"
 		},
-		"status": { "command": "rpm --quiet --query vscode" }
+		"status": { "command": "rpm --quiet --query code" }
```
bddenhartog: What about users who update Fedy with the `copr` repository installed? Now suddenly they won't see that Visual Studio Code is installed. This should probably query both temporarily -- this can be done by changing this command to `rpm --quiet --query code || rpm --quiet --query vscode`. You _do_ need the `or` separator, expanding both arguments (via shell expansion with curly braces) will not work properly. See my comment on `uninstall.sh` for more information.

plugins/visualstudiocode.plugin/uninstall.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-dnf copr -y disable  mosquito/vscode
-dnf -y --setopt clean_requirements_on_remove=1 erase  vscode -y
+dnf -y --setopt clean_requirements_on_remove=1 erase code
```
bddenhartog: Since changing the `status.command` value will show the correct state to users who had the `copr` repository installed, the `Remove` button will be displayed. This script should probably conditionally handle uninstalling both the `copr` and official repositories for now. After implementing this in your feature branch, you can make a follow up task to remove the `copr` repository conditionals for `vscode`.

plugins/visualstudiocode.plugin/install.sh
Artilex: That is indeed the case. However, icon themes that supply their own icon for vscode use the term `vscode` instead of `code`. Would you rather simply keep `icon=code` in the desktop file or rename the official icon from the repository whilst keeping `icon=vscode`?

bddenhartog: > However, icon themes that supply their own icon for vscode use the term vscode instead of code.

I would classify this as a bug within the icon theme itself, and users can either make the modifications themselves or report it to the theme's author through the proper channels. Fedy plugins should not assume things about the user's environment, like whether or not they are using an icon theme that doesn't support the `code` icon.

The real issue is that Microsoft didn't provide an official package in the past, so different users used different packages, which -- like the `mosquito` copr repository -- added binaries/icons/etc with the author's preferred names and structure, which causes differences in environments between users.

Artilex:
I'm at a bit of a loss regarding how the Install/Remove button is determined. The only way I've been able to set it to the desired state is by running a shell script from within `"status": {}` which deliberately returns an error when neither package is installed and does "doesn't-matter-what" when one is:

```shell
if command rpm -q --quiet code || rpm -q --quiet vscode; then
    # One of the packages is installed
    :
else
    # Neither is installed
    Deliberately-return-an-error
fi
```

Simply `"status": { "command": "rpm --quiet --query code || rpm --quiet --query vscode" }` or `rpm --quiet --query code vscode` does not set the button to the desired state in any case.

Spitting through `app.js` didn't make me much wiser.

Any ideas?

bddenhartog:
@Artilex Oops, my bad, I didn't check wrote I wrote in the application. We determine the plugin's status [here](https://github.com/folkswithhats/fedy/blob/master/app.js#L326-L344), by executing the value of `scripts.status.command` within each plugin's `metadata.json` file.

It's possible the `or` operator isn't being parsed correctly. Can you try escaping each pipe character and let me know the result?

```bash
"command": "rpm --quiet --query code \|\| rpm --quiet --query vscode"
```

plugins/visualstudiocode.plugin/metadata.json
bddenhartog: The `status.command` value can be set to `sh -c 'rpm --quiet --query code || rpm --quiet --query vscode'` -- sorry for the invalid (first) suggestion!

Artilex: Feeling a tad stupid for not thinking about that. It works and is quite a bit nicer than my solution which points to a script with:

```shell
if command rpm -q --quiet code || rpm -q --quiet vscode; then
	:
else
	Deliberately-return-an-error &> /dev/null
fi
```

bddenhartog: No worries! It wasn't the first thing I jumped to, either. 🙃

plugins/visualstudiocode.plugin/install.sh
```diff
@@ -1,4 +1,16 @@
 #!/bin/bash
 
-dnf copr -y enable mosquito/vscode 
-dnf -y install vscode
+gpgkey=$"https://packages.microsoft.com/keys/microsoft.asc"
```
bddenhartog: I think this `$` is erroneous - this concatenates with the string and would make `rpm --import` try to import `$https://packages.microsoft.com/keys/microsoft.asc`, which isn't a valid URI.

bddenhartog:
@Artilex are you able to make that minor change?

@satya164 Can you please review and approve your existing review?

Issue #439, opened 2017-03-07
Readme needs updated installation instructions


RafalSkolasinski:
A installation [script](https://www.folkswithhats.org/fedy-installer) seems unavailable
```
An error occurred.

Sorry, the page you are looking for is currently unavailable.
Please try again later.

If you are the system administrator of this resource then you should check the error log for details.

Faithfully yours, nginx.
```

bddenhartog:
Where are you encountering this? The script is simply named `installer`, not `fedy-installer` like the URI you linked to. Follow the installation instructions on www.folkswithhats.org should result in a successful installation.

RafalSkolasinski:
seems like link in readme: https://github.com/folkswithhats/fedy/blob/master/README.md#installation is out of date

bddenhartog:
Thanks @RafalSkolasinski. I'll get the readme updated shortly (really, it should just direct users to the website).

RafalSkolasinski:
No problem!

PR #451, opened 2017-04-12
update installation instructions in readme

bddenhartog:
- remove the invalid link to the installation script
- direct users to the website, and emphasize that no other source should
be used for installing Fedy

closes gh-439


bddenhartog:
Sorry 'bout that, a script I'm working on published this PR for an old branch. As an aside, it's interesting that a diff is shown considering this commit already exists in `master`.

Issue #452, opened 2017-04-24
Refactor Popcorn Time to source the CE version from Github


guinuxbr:
According to this: http://popcorn-time.is/official-statement.html#Malware2, the PopcornTime installer that is available at popcorntime.sh is a threat. Maybe the URL should be replaced by the PopcornTime CE.
This is the actual Popcorn Time URLs.
http://popcorn-time.is/
https://github.com/PopcornTimeCommunity/desktop

bddenhartog:
I'm considering this a **high priority** and will resolve it shortly.

bddenhartog:
Patch submitted for review, but before we merge this in, I have a few questions.

- Why is `PopcornTimeCommunity` more trustworthy than `popcorn-official`? Both applications are open-sourced:
    - https://github.com/PopcornTimeCommunity/desktop
    - https://github.com/popcorn-official/popcorn-desktop
- [This image](https://i.imgur.com/cRBrr6z.png) from the subreddit [/r/PopCornTime](https://reddit.com/r/PopCornTime) seems to directly contradict the image displayed in the blog post you linked. What gives?

bddenhartog:
Additionally:

* I ran [a scan](http://archive.is/lLSZQ) with `virustotal.com` with the download provided by `popcorntime.sh` for Windows 7+. There are no scanners that reported the file as unsafe, including Symantec (the only scanner that reported 'unsafe' on the file in the scan linked in the article).

* The file name in the blog post's `virustotal.com` scan is `Popcorn-Time Official 2016-Setup-ByINFORMACIONLIBRE.exe`, which is not the file name of the download provided by `popcorntime.sh`.

In other words, I can't tell that the information in the article you linked is accurate. It looks clean by all accounts.

guinuxbr:
Hi... After a little more research I can assume that you are right. There's no easy way to determine which domain is real or fake. They all claim to be clean. I'm sorry for this.

PR #453, opened 2017-04-25
refactor popcorn time to install popcorn time CE (yify)

bddenhartog:
closes gh-452


bddenhartog:
Some notes:

* This plugin has been refactored to install (and test for installation) on a per-user basis, so it supports Fedy being installed in multi-user environments
* `uninstall.old.sh` is executed with `run-as-root` so that the previous version of the plugin (which installed the program globally) can be removed successfully

bddenhartog:
Per the conversation in #452, I'll be refactoring this to source the previous origin - `popcorntime.sh`.

PR #454, opened 2017-05-03
Fixes #449 Regarding Smartgit.

### LINKED ISSUE #449: SmartGit Plugin
Dear Fedy devs,

I have noticed that the SmartGit plugin was broken in Fedy 4.2.5. 

I have fixed it the best I can. 

1. I fixed the FILE extraction: (head+grep+tr) -> sed
2. I fixed the URL format

```sh
FILE=smartgit-linux-$(wget "https://www.syntevo.com/smartgit/changelog.txt" -O - |  sed -n '1{s/SmartGit \([0-9][0-9]*\)\.\([0-9][0-9]*\)\.\([0-9][0-9]*\) .*/\1_\2_\3/ ; p}').tar.gz
URL=http://www.syntevo.com/static/smart/download/smartgit/$FILE
```

It works at least for me.
Cheers

Abhinav1217:
Tested and reviewed the link. It is working as expected.


plugins/smartgit.plugin/install.sh
```diff
@@ -6,8 +6,8 @@ CACHEDIR="/var/cache/fedy/smartgit-generic";
 mkdir -p "$CACHEDIR"
 cd "$CACHEDIR"
 
-FILE=smartgit-generic-$(wget "https://www.syntevo.com/smartgit/changelog.txt" -O - | head -n 1 | grep -Po '(?<=SmartGit )\d.\d.\d' | tr "." _).tar.gz
-URL=http://www.syntevo.com/downloads/smartgit/$FILE
+FILE=smartgit-linux-$(wget "https://www.syntevo.com/smartgit/changelog.txt" -O - |  sed -n '1{s/SmartGit \([0-9][0-9]*\)\.\([0-9][0-9]*\)\.\([0-9][0-9]*\) .*/\1_\2_\3/ ; p}').tar.gz
```
bddenhartog: * The beginning of your regular expression should be rewritten to `s/^.*` since we don't want to assume that they will always have `SmartGit` as the exact and only text prior to the version number.

* The groups inside your regular expression can be rewritten as `\([0-9]\+\)\.\([0-9]\+\)\.\([0-9]\+\)`, which is a fair bit less verbose and still matches on three groups of 1 of more digits separated by a space. It's unfortunate we can't use the (even shorter) `\(\([0-9]\+\.\?\)\{3\}`, which would match the entire version in a single group -- simply because we need to replace the periods with underscores. Maybe this (replacing the periods with underscores) should be handled in a separate, second expression?

Abhinav1217: Its the same code given in #449 . Also It is an immediate fix because the old plugin is not working at all.  For long term fix, I would actually prefer to look for download link from the[ download page itself.](https://www.syntevo.com/smartgit/download). Like we do for other softwares.
 
Regarding Point 1.
They will always use SmartGit as exact word because thats how they have stylised their app name.  They have a fixed format for their [Changelog](https://www.syntevo.com/smartgit/changelog.txt).  
This script targets the Stable release.  I couldn't test for `SmartGit 17 RC 3 (2017-01-12)` download because they didn't exists. 

Regarding point two. Yes it is less verbose, and serves same result. There there is no added advantage to use it.  Less verbose means harder to understand and debug.

PR #455, opened 2017-05-13
Add Arc Icon Theme

### LINKED ISSUE #392: Arc Icon Theme
Add please Arc icon theme
https://github.com/horst3180/arc-icon-theme

user501254:
A flat icon theme to accompany the Arc GTK theme.

Makes use of [this Copr Repository](https://copr.fedorainfracloud.org/coprs/user501254/Arc/).
On commits, SRPM packages are submitted to Copr on fixed intervals, the script and spec file can be found in [this repository](https://github.com/user501254/packaging).
The Copr repository also provides `moka-icon-theme` & `faba-icon-theme` which is used by the `arc-icon-theme` in case of missing application icons.

This would close #392.

### DIFF
--- a/plugins/arcicontheme.plugin/metadata.json
+++ b/plugins/arcicontheme.plugin/metadata.json
@@ -1,0 +1,17 @@
+{
+	"label": "Arc Icon Theme",
+	"description": "A flat icon theme to accompany the Arc GTK theme.",
+	"license": "GPLv3",
+	"category": "Themes",
+	"scripts": {
+		"exec": {
+			"label": "Install",
+			"command": "run-as-root -s install.sh"
+		},
+		"undo": {
+			"label": "Remove",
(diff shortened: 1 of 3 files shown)


plugins/arcicontheme.plugin/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+	"label": "Arc Icon Theme",
+	"description": "A flat icon theme to accompany the Arc GTK theme.",
```
bddenhartog: This sort of reads like you _need_ "Arc GTK Theme" installed. This might be confusing to new users; can we re-word this to simply `A flat icon theme`, or something that makes it obvious that this can be installed on its own?

user501254: There's no official "description" for the icon theme as of yet in the project's README file. But I got you point, will update.

PR #458, opened 2017-07-07
Update install.sh on VirtualBox plugin

bheru27:
Just tried to install the thing and the package is called VirtualBox-5.1 example: VirtualBox-5.1.x86_64 : Oracle VM VirtualBox, anyway, package VirtualBox exists too since fedy adds the rpmfusion repos.


bddenhartog:
Are the packages `Virtualbox` (from RPMFusion) and `Virtualbox-5.1` (from the standard repositories) on the same / effectively similar release schedule?

If so, it would make sense to use RPMFusion's package to avoid having to update this on each new release.

bddenhartog:
Note that PR #461 proposes using the RPMFusion repositories.

bheru27:
Yeah it *is* the same version from the rpmfusion repo

bddenhartog:
Alright, I'm going to close this in preference of the change bundled in #461 then.

PR #462, opened 2017-07-12
Some new packages

Abhinav1217:
Added Google Play, Cerebro, and Yatta Launcher, Peazip and wxHexEditor


plugins/googlemusic.plugin/uninstall.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+dnf -y --setopt clean_requirements_on_remove=1 remove google-play-music-desktop-player
```
bddenhartog: Move this to `metadata.json` and delete this file. Additionally, remove the`setopt` portion of the command.

plugins/peazip.plugin/uninstall.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+dnf copr -y disable brollylssj/PeaZip
+dnf -y --setopt clean_requirements_on_remove=1 remove peazip
```
bddenhartog: Remove the `setopt` portion of the command. The option defaults to True.

plugins/wxhexeditor.plugin/uninstall.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+dnf copr -y disable brollylssj/wxHexEditor
+dnf -y --setopt clean_requirements_on_remove=1 remove wxHexEditor
```
bddenhartog: `clean_requirements_on_remove` defaults to `True`. Remove this portion of the command.

plugins/yatta.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [[ -f /opt/yatta-eclipse-launcher.jar ]]; then
```
bddenhartog: This conditional check is unnecessary with the `-f|--force` option on `rm`.

plugins/yatta.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
...
+  rm -rf /opt/yatta-eclipse-launcher.jar
```
bddenhartog: This can be moved to `metadata.json` and this file removed.

plugins/cerebro.plugin/install.sh
```diff
@@ -0,0 +1,19 @@
...
+URL=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/KELiON/cerebro/releases/latest)
+
+DLF=https://github.com/KELiON/cerebro/releases/download/${URL##*/}/cerebro-$( echo ${URL##*/} | cut -c 2-)-x86_64.AppImage
+
+wget -c $DLF -O cerebro.AppImage
+
+chmod +x cerebro.AppImage
+mv cerebro.AppImage /opt/cerebro.AppImage
+
+# first run needed to integrate in system.
```
bddenhartog: So this doesn't actually set it up for the user, and they need to manually run a command?

It would seem, then, like this shouldn't be included in this PR.

plugins/cerebro.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [ -f /opt/cerebro.AppImage ]; then
```
bddenhartog: This conditional check is unnecessary with the `-f|--force` option on `rm`.

plugins/cerebro.plugin/uninstall.sh
```diff
@@ -0,0 +1,5 @@
...
+  rm -f /opt/cerebro.AppImage
```
bddenhartog: This can be moved to `metadata.json` and this file removed.

plugins/cerebro.plugin/uninstall.sh
Abhinav1217: Does't the `-f` in the if condition different from the `-f` argument in rm command.

in if condition, it checks for if the path is a file or not. and `rm -f` is force remove.

Why this is unnessary? It will give error.

plugins/cerebro.plugin/uninstall.sh
Abhinav1217: What's wrong with a proper separate uninstall file? Separation of concern is good practice.

plugins/cerebro.plugin/install.sh
Abhinav1217: yes, this is an issue. as fedy is executed as root user, and GUI apps launching as different user will cause issue ' no Protocol found.`

I have found a workaround I am trying to get it resolved.

You can skip this.

bddenhartog: It can't be "skipped". You need to remove this from the PR.

plugins/cerebro.plugin/uninstall.sh
bddenhartog: The `-f` conditional checks that the argument is a file, yes.

The `-f` _option_ for the `rm` command is the shorthand form of the option `--force`. With this option, `rm` will always have an exit code of `0`, regardless of whether or not the file exists ([source](
http://man7.org/linux/man-pages/man1/rm.1.html)).

If you were experiencing failures without the conditional check, they were likely due to something else.

bddenhartog: FYI, you can test this in a terminal by executing the following:

```bash
$ touch /tmp/somefile

$ rm -f /tmp/somefile

$ echo $?
> 0

$ rm -f /tmp/somefile

$ echo $?
> 0

Issue #448, opened 2017-03-16
[RFC] Building plugins as system packages instead of shipping shell scripts


sudoforge:
# Overview

Currently, the majority of plugins supported by Fedy and the Folks With Hats team are shipped as shell scripts, utilizing the `install.sh`, `uninstall.sh`, and `metadata.json` scripts. As a result of this decision, there are several points of pain with the current application workflow.

Note that in regards to #422, #447, this proposal should be considered made for `officially supported plugins` only. Community-built plugins are still a WIP, and we do not have a specific plan for implementation.

## Definitions

To start, let's define a few terms so that everyone is on the same page when discussing this topic.

- `target` is the resulting application, tool, or other item(s) that ends up on the user's system, after installing dependencies and configuration is completed.
- `plugin` means the bundle of commands/scripts used by Fedy to install a `target`.
- `package` is an RPM package, built from a `spec` file.

## Problems with scripts

There are several problems with the current way we ship plugins.

### Confusing upgrade steps

Once a user installs a target through Fedy's plugin workflow, there is no way to update the steps defined within the plugin. If we need to ship changes to a user, like say, a refactored installation step or new version of the target, they need to:

- uninstall the target by clicking the `Remove` button
- update the `fedy` application via `dnf up{date,grade}` to get the latest bundled versions of the plugin's scripts
- reinstall the target by clicking the `Install` button.

Additionally, these are the same steps a user would have to take in order to update a target that Fedy installed, but which doesn't have a built-in method for updating.

### Unmanaged versions

Many plugins install the latest version of the target from the target's origin source. This can be convenient, but allows for inconsistencies between users of Fedy, which can sometimes be the source of a bug, and can certainly cause issues when trying to replicate an issue. If a user installed a target from a plugin in January, that target could receive weekly updates, and the user would be unaware. They might report a bug six months down the road, and due to unmanaged versions of targets, it may be more difficult to track down that bug.

## Pros of using system packages 

- All plugins are easily manageable, and updates can be shipped to users via their next system update
- Targets can be locked into specific versions, which will ensure that all plugins of a specific release will always work as expected
- Simpler file tree: a single `spec` file can be used to define all attributes and (un/in)stallation steps currently split across `install.sh`, `uninstall.sh`, and `metadata.json`.
[...]

PR #472, opened 2017-07-19
Review of existing plugins

kwizart:
This is a partial review of existing plugins to verify compatibility and avoid mistake and possible error.

From this review:
- Remove dependencies with conflicting repository
- Fix VirtualBox module, one cannot install the guest and hyp on the same host
- Use a dedicated plugin for flash ppapi (chromium based)
- Rely on multimedia group to have a more accurate codecs installation

Looking at the way repo/software are dealt with, I think it should be better to have a dedicated plugin for each repository (copr, *-release or else). Then have software to add a dependency on the repo plugin. If 

There are also lot of binary software that are installed into /opt whereas they could be installed into a user specific directory. It's not a big issue, but those would be candidates for flatpak...

### DIFF
--- a/plugins/flash.plugin/metadata.json
+++ b/plugins/flash.plugin/metadata.json
@@ -1,7 +1,7 @@
 {
 	"icon": "flash",
-	"label": "Adobe Flash",
-	"description": "Browser plug-in and rich Internet application runtime.",
+	"label": "Adobe Flash (npapi)",
+	"description": "Browser plug-in and rich Internet application runtime for npapi enabled browser (firefox, etc)",
 	"license": "Proprietary",
 	"category": "Utilities",
 	"scripts": {
--- a/plugins/flash.plugin/metadata.json
+++ b/plugins/flash.plugin/metadata.json
@@ -11,7 +11,7 @@
 		},
 		"undo": {
 			"label": "Remove",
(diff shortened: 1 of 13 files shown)


plugins/flash.plugin/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
 	"icon": "flash",
-	"label": "Adobe Flash",
-	"description": "Browser plug-in and rich Internet application runtime.",
+	"label": "Adobe Flash (npapi)",
```
kwizart: Not exactly. Or at least it's on purpose. The npapi is the historic version of the (adobe) flash plugin. and the package name is the same. (It has seen a major version bump after the plugin was deprecated for a while).
So the change here is only to improve the current plugin description.

In the fedy context, I expect it's a bad thing to rename plugins, because it might be considered a new item not replacing the old plugin. This is specially important if you want to enable the browser mode where plugins will be fetched from github. IMO, I don't think this browser mode feature is very important over distributing the full plugins using RPM.

bddenhartog: The FWH organization doesn't really have guidelines surrounding plugin renaming, but I can understand your reasoning based on your other points. 

Regarding RPM distribution of plugins, I wholeheartedly agree with you. See #448.

PR #478, opened 2017-07-27
Install libXScrnSaver with GitKraken

Dshoe:
libXScrnSaver is a required dependency that is not installed by default on Fedora 26. If this library is not present then GitKraken will fail to start with the error: "error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory"


Dshoe:
Thanks for the merge!

PR #484, opened 2017-08-28
Add Brave Browser and other fixes #482

### LINKED ISSUE #482: Add Brave Browser?
Would there be any interest or would the project be able to add the Brave Browser to FEDY? 

https://github.com/brave/browser-laptop/blob/master/docs/linuxInstall.md#fedora-x86_64

Abhinav1217:
This will add Brave browser in plugin directory.
Also fixes an unreported bug in sublime text 3 installation.

Closes #482

### DIFF
--- a/plugins/brave.plugin/install.sh
+++ b/plugins/brave.plugin/install.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Instructions adopted from
+# https://github.com/brave/browser-laptop/blob/master/docs/linuxInstall.md#fedora-x86_64
+
+dnf config-manager --add-repo https://s3-us-west-2.amazonaws.com/brave-rpm-release/x86_64/
+rpm --import https://s3-us-west-2.amazonaws.com/brave-rpm-release/keys.asc
+dnf -y install brave
--- a/plugins/brave.plugin/metadata.json
+++ b/plugins/brave.plugin/metadata.json
@@ -1,0 +1,19 @@
+{
+	"icon": "brave",
+	"label": "Brave Browser",
+	"description": "Browse faster by blocking ads and trackers",
+	"license": "Mozilla Public License 2.0",
+	"libre": true,
+	"category": "Apps",
+	"scripts": {
+		"exec": {
+			"label": "Install",
+			"command": "run-as-root -s install.sh"
+		},
+		"undo": {
+			"label": "Remove",
+			"command": "dnf -y remove brave"
+		},
+		"status": { "command": "rpm --quiet --query brave" }
+	}
+}
--- a/plugins/brave.plugin/metadata.json
+++ b/plugins/brave.plugin/metadata.json
@@ -1,7 +1,7 @@
 {
 	"icon": "brave",
 	"label": "Brave Browser",
-	"description": "Browse faster by blocking ads and trackers",
+	"description": "A privacy focused browser that blocks ads and trackers by default",
 	"license": "Mozilla Public License 2.0",
 	"libre": true,
 	"category": "Apps",
@@ -12,7 +12,7 @@
 		},
 		"undo": {
 			"label": "Remove",
-			"command": "dnf -y remove brave"
+			"command": "run-as-root dnf -y remove brave"
 		},
 		"status": { "command": "rpm --quiet --query brave" }
 	}
--- a/plugins/sublimetext.plugin/install.sh
+++ b/plugins/sublimetext.plugin/install.sh
@@ -2,4 +2,4 @@
 
 rpm -v --import https://download.sublimetext.com/sublimehq-rpm-pub.gpg
 dnf config-manager --add-repo https://download.sublimetext.com/rpm/stable/x86_64/sublime-text.repo
-dnf install sublime-text
+dnf -y install sublime-text


plugins/brave.plugin/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Instructions adopted from
```
bddenhartog: For future reference, this comment is something that could be put in the commit's body, so that it becomes more easily searchable, and visible to other users.

plugins/brave.plugin/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
+	"icon": "brave",
+	"label": "Brave Browser",
+	"description": "Browse faster by blocking ads and trackers",
+	"license": "Mozilla Public License 2.0",
+	"libre": true,
```
bddenhartog: Based on [what I can find](https://github.com/folkswithhats/fedy/search?q=libre&type=Code&utf8=%E2%9C%93), it looks like the `license` field always wins over `libre`.

Can you help explain to me what the differences are between them, and how a developer should choose between providing either the `license` or `libre` properties, or both?

plugins/brave.plugin/install.sh
bddenhartog: Just FYI: Code comments like this might be better off being put in the body of the commit when the relevant code is added.

plugins/brave.plugin/metadata.json
```diff
@@ -0,0 +1,19 @@
...
+	"category": "Apps",
+	"scripts": {
+		"exec": {
+			"label": "Install",
+			"command": "run-as-root -s install.sh"
+		},
+		"undo": {
+			"label": "Remove",
+			"command": "dnf -y remove brave"
```
bddenhartog: Since we're calling dnf, don't we need to prefix the command with `run-as-root`? The only way this would work as it is now is if the user opened fedy as the root user.

plugins/sublimetext.plugin/install.sh
```diff
@@ -2,4 +2,4 @@
 
 rpm -v --import https://download.sublimetext.com/sublimehq-rpm-pub.gpg
 dnf config-manager --add-repo https://download.sublimetext.com/rpm/stable/x86_64/sublime-text.repo
-dnf install sublime-text
+dnf -y install sublime-text
```
bddenhartog: Good catch! Ideally this would be done in its own branch, with its own PR, so it didn't get held up by the PR in case additional work or clarification is needed and it can't go in right away.

This PR is small enough to not worry about that though. **Thank you** for making this change in its own commit! 👍

plugins/brave.plugin/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
+	"icon": "brave",
+	"label": "Brave Browser",
+	"description": "Browse faster by blocking ads and trackers",
```
bddenhartog: Can we reword this to something along the lines of

> Brave is an open source, privacy-and-performance focused browser that blocks ads and trackers by default

Check the length of that, and adopt as needed!

plugins/brave.plugin/metadata.json
Abhinav1217: `License` field contains string, or array of string which will be displayed beside the name of plugin.  `Libre`, `propriety`, `patented` and `mixed` were some fallbacks kind of thing, in case license is not clear or there are multiple license. `has_license` condition on line 552 in app.js might shed light on how it works now. 

License was the first thing we discussed after we formed this org on telegram. Search for license on telegram group, @satya164  worked on this between 12-08-15 and 16-08-15.   

Original issue was #211 and PR was https://github.com/folkswithhats/fedy/commit/1d0421b7fc53d08b1618b312e0f2b9f4a6b3c9b4#diff-0364f57fbff2fabbe941ed20c328ef1aR552

plugins/sublimetext.plugin/install.sh
Abhinav1217: It was really small change that's why I didn't think of seperate PR for this. But I will keep individual changes on separate commit and group them in separate branches just like you taught me last time.

plugins/brave.plugin/metadata.json
Abhinav1217: Somehow I mixed that up in commit. Fixing it now.

plugins/brave.plugin/metadata.json
Abhinav1217: That was picked from their site directly. I will rephrase it and send PR.

plugins/brave.plugin/install.sh
Abhinav1217: I started keeping reference links and comments in the file as docs for other contributors working on it.  I will start adding links it to the commit messages too if you want.  I didn't add do that at first because these links are most of the time provided and discusses in PR messages and Issue messages too.

plugins/brave.plugin/metadata.json
bddenhartog: Right - I understand what the initial intent was, but currently, the `libre` property is not used anywhere:

https://github.com/folkswithhats/fedy/blob/master/app.js#L530-L539

The lines that you'll see highlighted, L530 - L539, are the only lines related to license display that I'm able to find. They don't use the `libre` property at all.

Abhinav1217: @satya164 would be better explain this as he was the one who worked on it. As far as I know, these tags was supposed to be fallback for when License was not known or not clear. Then these tags could be used to simply convey, this app is free to use, this app is propriety, etc. If license field is there, It would be used. 

Let's take this to telegram, and when things are clear, we will add it to repo documentation.

bddenhartog: > Let's take this to telegram, and when things are clear, we will add it to repo documentation.

That causes visibility to be lost to the users.

I suppose having this superfluous property isn't going to harm anything, and we can later do a PR to remove all of them, or re-implement support. I'll open an issue tonight.

Issue #486, opened 2017-08-30
The `libre` property in plugin metadata is unused and should be removed


sudoforge:
Initial discussion here: https://github.com/folkswithhats/fedy/pull/484#discussion_r135539992

There are [several plugins](https://github.com/folkswithhats/fedy/search?utf8=%E2%9C%93&q=libre&type=) with the superfluous `libre` property. Go through the codebase and remove these properties.

Quick-start in your terminal/shell (if you're working from a fork, be sure to update your fork first):

```
$ git checkout master
$ git pull
$ git checkout -b feature/486/remove-libre-property master
```

PR #488, opened 2017-09-01
[WIP] Expand Software Offering: Anydesk

89luca89:
Add Anydesk Application


plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,14 @@
...
+VERSION=$(wget https://anydesk.com/download -O - | grep -e "v.[0-9].[0-9].[0-9]" | head -n 1 | sed 's/<small>//' | sed 's/<\/small><br>//' | sed 's/ //g' | sed 's/\t//g')
+if (( $? == 0));then
+	# 64 bit
+	URL="https://download.anydesk.com/linux/rhel7/anydesk-"$VERSION"-1.el7.x86_64.rpm"
+else
+	# 32 bit 
+	URL="https://download.anydesk.com/linux/rhel7/anydesk-"$VERSION"-1.el7.i686.rpm"
+fi
+wget $URL -O "/var/cache/fedy/anydesk.rpm"
+dnf -y install "/var/cache/fedy/anydesk.rpm"
```
bddenhartog: Should be `/var/cache/fedy/anydesk/anydesk.rpm`

plugins/anydesk.plugin/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
+	"icon":"anydesk",
+	"label": "AnyDesk",
+	"description": "AnyDesk is the world's most comfortable remote desktop application. Access all your programs, documents and files from anywhere",
+	"license": "Proprietary",
+	"libre": false,
```
bddenhartog: This property is not used anywhere in the application.

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+
+arch | grep x86_64
+VERSION=$(wget https://anydesk.com/download -O - | grep -e "v.[0-9].[0-9].[0-9]" | head -n 1 | sed 's/<small>//' | sed 's/<\/small><br>//' | sed 's/ //g' | sed 's/\t//g')
```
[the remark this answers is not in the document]
bddenhartog: The `wget` command here is a bit fragile. Right now, it works because all of their `v.X.X.X` tags are at the same version. What happens when it's not?

Try looking for and parsing the full url for the fedora version we're in, instead:

```
https://download.anydesk.com/linux/fedora26/anydesk-2.9.4-1.fc26.x86_64.rpm
```

plugins/anydesk.plugin/metadata.json
89luca89: I used as template some other Metadata files.
Look at Atom plugin for example

plugins/anydesk.plugin/install.sh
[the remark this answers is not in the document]
89luca89: Anydesk is not released for Fedora 26
So I think using the Rhel7 package was wiser than using a F24 one

Any suggestion on what to do?

screenshot from 2017-09-02 17-20-13

plugins/anydesk.plugin/metadata.json
bddenhartog: @89luca89 Yes, I'm aware that it exists as a property in a few other plugins. This is erroneous, as the support for displaying this property was removed from the GUI application some time ago. See https://github.com/folkswithhats/fedy/issues/486.

plugins/anydesk.plugin/install.sh
```diff
@@ -1,8 +1,14 @@
-#!/bin/bash
+#!/bin/env bash
```
bddenhartog: Use `/usr/bin/env bash`.

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,24 @@
...
```
bddenhartog: Use `/usr/bin/env bash`.

plugins/anydesk.plugin/install.sh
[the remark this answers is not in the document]
bddenhartog: @89luca89 Sorry, this comment got buried somehow and I'm just now seeing it. Hmm... I see the problem now. Unfortunately, there's not much we can do. I suppose the RHEL7 release is fine, and shouldn't cause problems. I haven't done any actual box testing yet for this (or any of the swath of PRs), just code review. If there's an issue with the RHEL7 version, I'll catch it during box tests.

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,24 @@
...
+VERSION=$(wget https://anydesk.com/download -O - |\
+    grep -e "v.[0-9].[0-9].[0-9]" |\
+    head -n 1 |\
+    sed 's/<small>//' |\
+    sed 's/<\/small><br>//' |\
+    sed 's/ //g' |\
+    sed 's/\t//g')
+
+arch | grep x86_64
+if (( $? == 0));then
```
bddenhartog: `if` clause can be simplified to:

```
if command arch | command grep 'x86_64'; then
    ...
fi
```

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,24 @@
...
+
+VERSION=$(wget https://anydesk.com/download -O - |\
+    grep -e "v.[0-9].[0-9].[0-9]" |\
+    head -n 1 |\
+    sed 's/<small>//' |\
+    sed 's/<\/small><br>//' |\
+    sed 's/ //g' |\
+    sed 's/\t//g')
+
+arch | grep x86_64
```
bddenhartog: Remove this line.

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+CACHEDIR="/var/cache/fedy/Anydesk";                                                                                                                           
+mkdir -p "$CACHEDIR"    
+cd "$CACHEDIR"
+
+VERSION=$(wget https://anydesk.com/download -O - |\
+    grep -e "v.[0-9].[0-9].[0-9]" |\
```
bddenhartog: My earlier review comment regarding this `grep` command wasn't clear enough - my apologies.

This is fragile because it relies on their version numbers always being one digit - like `v.1.2.3`. This will fail when any of the major/minor/patch numbers reach double digits, e.g. `v.1.2.11`. You should use the `+` operator to require _one or more_ of the preceding terms. Additionally, you can simplify this and cut down on repetition by using a regex group and repeater.

This is a more flexible regex pattern implementing the above comments:

```
grep -e 'v.\([0-9]\+\.\?\)\{3\}'
```

plugins/anydesk.plugin/install.sh
```diff
@@ -0,0 +1,24 @@
...
+    head -n 1 |\
+    sed 's/<small>//' |\
```
bddenhartog: These `sed` commands can be simplified into a single expression:

```
sed 's/.*\(v\.\([0-9]\+\.\?\)\{3\}\).*/\1/'
```

Instead of matching on undesirable characters and removing them, this matches on the specified pattern (which matches the `grep` command from earlier) and ignores everything else.

PR #489, opened 2017-09-01
[WIP] Expand Software Offering: Bleachbit

89luca89:
Add BleachBit Application


bddenhartog:
`bleachbit` is available in the standard Fedora repository,

89luca89:
This , like also others , was included to give a graphical way to install some important software that are not present in the Store (but still present in standard repository)

bddenhartog:
@89luca89 I totally understand - however, the current maintainers recently made the decision that plugins for packages available from the standard repositories is something we want to avoid.

Issue #485, opened 2017-08-28
Master PDF editor does not download


gitgt:
Master PDF editor does not seem to download. Console shows:
http://get.code-industry.net/public/master-pdf-editor-4.2.70_qt5.x86_64.rpm
and stays like that. Manual download works.

danstewart:
Fix merged #494

rishadfb:
It's broken again. Doesn't install for me.

sudoforge:
@rishadfb I've created a new issue for this problem: #592

PR #494, opened 2017-09-01
Fix MasterPdfEditor Install

89luca89:
Fix URL Creation for Master PDF Editor


danstewart:
Just linking to the original issue: #485

PR #508, opened 2017-09-01
[WIP] Expand Software Offering: LMMS

89luca89:
Add LMMS Application


89luca89:
This, like others I've included, are not in "Software" but still in standard repo
It was intended to give a graphical way to install some important software

kwizart:
Then it means it's missing an appdata file, so best is to submit an appdata to lmms upstream.

PR #509, opened 2017-09-01
[WIP] Expand Software Offering: Mendeley

89luca89:
Add Mendeley Desktop Application


89luca89:
Ok we can wait and see if an rpm will be provided

Abhinav1217:
@kwizart  RPMs from developer are always preferred. Thanks for filing request on rpmfusion bugzilla. We have in past included apps which did not have RPM provided by the Developers. Sublime text, Brackets, and Atom, All of them used to have only deb packages and .tar.gz packages and only recently started providing official rpms. 

@89luca89  Thanks for the PR.  I am solely judging based on 2 factors, "App is propriety with locked eco-system" and "Is designed for a very targeted ( hence very limited ) user group".  Since we currently haven't completed APIs, and all the plugins are bundled with every release, we have to be selective of plugins to include. With all the other giants marketing students and universities, I don't think this will be missed if not included in fedy.  I do hope rpmfusion peoples manage to include it in their repos. Its up to other fedy members to decide on this.

PR #518, opened 2017-09-01
[WIP] Expand Software Offering: Pixeluvo

89luca89:
Add Pixeluvo Application


Abhinav1217:
Do we need this when there are multiple free and open source ( even free to use apps ) in this category? Does this offer something really unique over them.

89luca89:
Just found it very intuitive and nice UI
Feel free to judge by yourself if including it or not

bddenhartog:
> Do we need this when there are multiple free and open source ( even free to use apps ) in this category? Does this offer something really unique over them.

Why would we want to limit the choices users have?

Abhinav1217:
There is no free to use version for this app. Feature list and UI does not provide anything unique compared alternatives like darktable. The trial version is also a limited feature trial. It is limited to 800x600px images only.

bddenhartog:
Statements that should _probably_ be considered when thinking about whether or not an application should make it into Fedy:

> * There is no free to use version for this app.
> * The trial version is also a limited feature trial. It is limited to 800x600px images only.

These statements lead to a pretty poor end user experience after installation. Of course, this stems from an assumption that users are not researching the application before clicking 'install', and thus the limited feature trial, and lack of of free version, will probably come as a surprise.

Statements that should _not_ be a factor in whether or not an application makes it into Fedy:

> * Feature list and UI does not provide anything unique compared alternatives

All I'm getting at is that applications are never _exactly_ the same, and we shouldn't reject an application for being similar to another - that prevents our users from feeling out the alternatives and finding what works for them.

That said, closing due to the aforementioned reasons.

PR #520, opened 2017-09-01
[WIP] Expand Software Offering: QOwnNote

89luca89:
Add QOwnNote Application


Abhinav1217:
We already have similar apps in fedy core. With similar functionalities. 
Does this app provide any benefits over our current offerings?

89luca89:
It has nice Markdown editing capabilities
Nexcloud/owncloud support
I can choose whatever folder to save so It can be used with Dropbox or Google Drive or any sync software
I use it with Dropbox and Plain.txt on android for example

Abhinav1217:
Fair enough.  I will begin testing this tomorrow and submit review after that.

PR #521, opened 2017-09-01
[WIP] Expand Software Offering: Shotcut

89luca89:
Add Shotcut Application


plugins/shotcut.plugin/install.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/shotcut";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+URL=$(wget "https://github.com/mltframework/shotcut/releases" -O - | grep tar.bz2 | head -n4 | sed 's/<a href=\"//g' | sed 's/\" rel=\"nofollow\">//g' | grep -v strong | head -n 1)
```
Abhinav1217: Use github.com/.../release/latest which will redirect to latest version of any app on github release page.  
Please refer to the GoogleMusic Plugin in Fedy master to see how this type of download has been handled.

89luca89: Fixed with your suggestion,
But cannot be exactly like in googlemusic plugin because of how Shotcuts names their releases

plugins/shotcut.plugin/install.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/shotcut";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+URL=$(wget "https://github.com/mltframework/shotcut/releases/latest" -O - | grep tar.bz2 | sed 's/<a href=\"//g' | sed 's/\" rel=\"nofollow\">//g'| grep -v strong | head -n 1)
+
+URL=$(echo "www.github.com"$URL | sed 's/ //g')
```
Abhinav1217: Remove "www. " because using it will not download package using HTTPS.

Abhinav1217: Using curl {url_effective} should be more reliable compared to `sed`. Any minute changes to github interface will break this. Whereas curl method will break only when github will change its api structure, for which github will notify all the developers of changes.

89luca89: Yeah but the {url_effective} will return:
`
https://github.com/mltframework/shotcut/releases/tag/v17.08
`
and the VER will be
`v17.08`

But the developer follows a different nomenclature:
`shotcut-linux-x86_64-170801.tar.bz2`

And most importantly it can change, for example for version v17.05 the file name is
`shotcut-linux-x86_64-170503.tar.bz2`

We do not know the last digit (03 in case of 17.05, 01 in  case of 17.08) So how do you suggest to retrieve it without using the HTML Parsing?

PR #522, opened 2017-09-01
[WIP] Expand Software Offering: Skype-Beta

89luca89:
Add Skyper for Linux Beta Application


89luca89:
Do you mean to modify the plugin that provides the "old" skype?

kwizart:
It's not the old skype at all (not in the master version of the fedy repo, at least)
it downloads the 5.4.0.1 RPM, built on 17th of July 2017, so the more recent (stable) version.

Abhinav1217:
@89luca89 You must have forked this repo Before F26 update. A lot of plugins were updated including Skype-Beta.  Google Music, etc. We do not have "old skype" plugin anymore. It was updated with Skype Beta "Same code as what you provide here"

89luca89:
Yeah forked before F26
Feel free to close all the duplicate work, no problem

bddenhartog:
@89luca89 What needs to happen here is you need to merge/rebase this branch on top of our `master`, and compare what you've provided here with the content in `/plugins/skype.plugin`. These provide the same application, but you may have differences which you think are better (at the very least, you provide an icon for Fedy's GUI to use -- we don't currently do that).

Please apply your changes to the existing skype plugin, and resubmit.

89luca89:
I'm seeing that your skype plugin is pulling the rpm directly instead of installing the repo and let DNF do the rest
Is there a reason to prefer the single rpm approach before I do a rebase?

Abhinav1217:
Looks like they have started providing the repos now. At the time of writing that plugin, They were only providing RPM, Repos were not ready then. Similar to VS-Code. Earlier, VS code was rpm only, now they are providing repos.

Let me test your plugin, if everything works, we can adopt it.

bddenhartog:
@Abhinav1217 

> Let me test your plugin, if everything works, we can adopt it.

I'd rather test against adding the repository, as that's a better solution than installing the RPM directly.

Abhinav1217:
That is what I am doing. I am testing script to add dnf repos and install skype using dnf. Comparing .repo file his plugin creates with one provided by microsoft.

PR #525, opened 2017-09-01
[WIP] Expand Software Offering: Spider Oak

89luca89:
Add SpiderOak One Application


plugins/spideroak.plugin/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+	"icon": "SpiderOakONE",
+	"label": "SpiderOak One",
+	"description": "Secure, reliable backup to protect your files from data loss and ransomware",
+	"license": "Freemium",
```
Abhinav1217: License for this app ( as per their official website ) can be categorised as propriety. Change it to "Propriety, Free with paid plans" to make it more clear.

Abhinav1217: Dependencies are not satisfied for gnome. 
```
[Abhinav@fedora-26 test] $ SpiderOak
Traceback (most recent call last):
  File "<string>", line 6, in <module>
  File "__main__.py", line 128, in <module>
  File "__main__SpiderOakONE__.py", line 183, in <module>
  File "__main__SpiderOakONE__.py", line 168, in main
  File "oak/Oak.py", line 22, in <module>
  File "PyQt4/QtGui.py", line 26, in <module>
  File "PyQt4/QtGui.py", line 17, in _bbfreeze_import_dynamic_module
ImportError: /opt/SpiderOakONE/lib/libz.so.1: version `ZLIB_1.2.9' not found (required by /lib64/libpng16.so.16)
[Abhinav@fedora-26 test] $ SpiderOakONE 
Traceback (most recent call last):
  File "<string>", line 6, in <module>
  File "__main__.py", line 128, in <module>
  File "__main__SpiderOakONE__.py", line 183, in <module>
  File "__main__SpiderOakONE__.py", line 168, in main
  File "oak/Oak.py", line 22, in <module>
  File "PyQt4/QtGui.py", line 26, in <module>
  File "PyQt4/QtGui.py", line 17, in _bbfreeze_import_dynamic_module
ImportError: /opt/SpiderOakONE/lib/libz.so.1: version `ZLIB_1.2.9' not found (required by /lib64/libpng16.so.16)
[Abhinav@fedora-26 test] $ 
```

89luca89: It was a problem with the URL Downloading an old version of SpiderOak

SpiderOak is self-contained does not need particular dependencies (or else dnf should have installed them)
It was a problem between the static libz in the package and the generic libz (1.2.11) in F26

bddenhartog: @Abhinav1217:
> License for this app ( as per their official website ) can be categorised as propriety. Change it to "Propriety, Free with paid plans" to make it more clear.

The term [Freemium](https://en.wikipedia.org/wiki/Freemium) is well-known and descriptive. There shouldn't be a problem with using that term to describe an application's payment model -- but I think that there's a different problem with both the initial usage here and with your change request.

The `license` field should simply show the license; not the payment structure chosen by the company/team behind the application. We currently do not support a "Payment Structure" property, and thus the only thing that should remain in the `license` field is the license itself -- in this case, "Proprietary".

plugins/spideroak.plugin/install.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/spideroak";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+arch | grep x86_64
```
bddenhartog: This is unnecessary.

plugins/spideroak.plugin/install.sh
```diff
@@ -0,0 +1,26 @@
...
+if (( $? == 0));then
```
bddenhartog: You can simply do this instead:

```
if command arch | grep 'x86_64'; then
   ...
fi
```

plugins/spideroak.plugin/install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+wget -c "$URL" -O "$FILE"
+
+if [[ ! -f "$FILE" ]]; then
+	exit 1
+fi
+
+dnf -y install $FILE
+
+
```
bddenhartog: Unnecessary blank line. We only want a LF (line feed) at EOF to prevent warnings in vim or any other editor which expects an EOL character at the end of the file.

plugins/spideroak.plugin/install.sh
```diff
@@ -0,0 +1,26 @@
...
+FILE="spideroak.rpm"
+
+wget -c "$URL" -O "$FILE"
+
+if [[ ! -f "$FILE" ]]; then
+	exit 1
+fi
+
+dnf -y install $FILE
+
```
bddenhartog: Unnecessary blank line. We only want a LF (line feed) at EOF to prevent warnings in vim or any other editor which expects an EOL character at the end of the file.

plugins/spideroak.plugin/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+	"icon": "SpiderOakONE",
+	"label": "SpiderOak One",
```
bddenhartog: "One" should be "ONE" to match the product name.

plugins/spideroak.plugin/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+	"description": "Secure, reliable backup to protect your files from data loss and ransomware",
```
bddenhartog: Missing period at the end of the description.

plugins/spideroak.plugin/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+	"icon": "SpiderOakONE",
+	"label": "SpiderOak ONE",
+	"description": "Secure, reliable backup to protect your files from data loss and ransomware",
+	"license": "Freemium
```
bddenhartog: Missing a quote and comma here.

Additionally, the license value should be "Proprietary"

PR #527, opened 2017-09-01
[WIP] Expand Software Offering: TLP Plugin for Laptops

89luca89:
Add TLP/Powertop tweak
This is important to Laptop/Mobile users to improve battery life

### DIFF
--- a/plugins/tlp.plugin/install.sh
+++ b/plugins/tlp.plugin/install.sh
@@ -1,0 +1,10 @@
+#!/bin.bash
+
+dnf -y install tlp tlp-rdw powertop smartmontools
+systemctl enable powertop.service
+systemctl enable tlp.service
+systemctl enable tlp-sleep.service
+tlp start true
+semanage fcontext -a -t ifconfig_var_run /run/tlp/lock_tlp
+restorecon -R -v /run/tlp/lock_tlp
+
--- a/plugins/tlp.plugin/install.sh
+++ b/plugins/tlp.plugin/install.sh
@@ -1,4 +1,4 @@
-#!/bin.bash
+#!/bin/bash
 
 dnf -y install tlp tlp-rdw powertop smartmontools
 systemctl enable powertop.service
--- a/plugins/tlp.plugin/uninstall.sh
+++ b/plugins/tlp.plugin/uninstall.sh
@@ -1,4 +1,4 @@
-#!/bin.bash
+#!/bin/bash
 
 tlp false
 systemctl disable powertop.service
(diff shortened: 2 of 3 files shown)


plugins/tlp.plugin/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin.bash
+
+dnf -y install tlp tlp-rdw powertop smartmontools
+systemctl enable powertop.service
+systemctl enable tlp.service
+systemctl enable tlp-sleep.service
+tlp start true
+semanage fcontext -a -t ifconfig_var_run /run/tlp/lock_tlp
```
kwizart: If this service cannot work without changing the SElinux context, please report a bug and add a comment to the bug here. If ever the SElinux context need to be changed in the next tlp release, you will create a bug by using an wrong context here.

plugins/tlp.plugin/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin.bash
```
kwizart: typo

plugins/tlp.plugin/uninstall.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin.bash
```
kwizart: again, #!/bin/bash

plugins/tlp.plugin/install.sh
89luca89: This fixes only a warning
Without it, TLP Works correclty but `Selinux Troubleshooter` will fire a notification warning each sleep/resume

I can remove the context change for now?

kwizart: You can remove it, but submitting a bugzilla to the tlp component is needed, applications aren't expected to fire selinux warning.

Abhinav1217: Is there any recommended Selinux policy regarding tlp and powertop services?

89luca89:
@kwizart  
If you are interested I have searched what specifically TLP/Powertop does and developed a udev rule

https://gist.github.com/89luca89/29be32da857fa4aa59b09ab626929d2f

This does 90% of what TLP does (and covers 100% of the powertop functionality)
Only missing things from TLP are the thinkpad specific functions (charge/discharge treshold) and CPU customization ([Here](https://gist.github.com/89luca89/b44731223311e3c4130d216c24dc2574) a script that can go with the udev rule switching AC/Battery)

Just with this rule the battery life of laptops can be improved dramatically

PR #528, opened 2017-09-01
[WIP] Expand Software Offering: uGet

89luca89:
Add uGet Application


Abhinav1217:
This app is available in gnome-software and KDE Discover, There is no point having it here.

89luca89:
I cannot find it using F26 in gnome-software

screenshot from 2017-09-04 11-30-13

bddenhartog:
@89luca89 I can confirm that the `uget` package does not appear in Gnome Software using the default settings. @Abhinav1217, are you sure that you can see it inside of Gnome Software?

Regardless, the `uget` package is available from Fedora's standard repositories, and as such this will be declined. Users who want this package can simply run `dnf install uget`.

PR #529, opened 2017-09-01
[WIP] Expand Software Offering: Whatever (Evernote)

89luca89:
Add Whatever (Evernote Client) Application


Abhinav1217:
I am comparing this with [Tusk](https://github.com/champloohq/tusk/) which appears to have more functionality than this. Will submit a review on which is better to add after a few week of using both of them.

PR #530, opened 2017-09-01
[WIP] Expand Software Offering: Whatsie

89luca89:
Add Whatsie (Whatsapp Client) Application


plugins/whatsie.plugin/install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+dnf -y install libappindicator-gtk*
+
+CACHEDIR="/var/cache/fedy/WhatsApp";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+URL="ftp://ftp.ntua.gr/pub/linux/pclinuxos/pclinuxos/apt/pclinuxos/64bit/RPMS.x86_64/whatsie-2.0.15-linux-x86_64.rpm"
```
kwizart: This package is 64bit (only) , but there is no architecture check.
I think the check should be done in the metadata in the first step.

Also, if it's only about one single rpm, it could be done directly in the metadata.

plugins/whatsie.plugin/install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+dnf -y install libappindicator-gtk*
```
kwizart: It is too much, it will also install development packages: (libappindicator-gtk3-devel)

89luca89: Will fix

plugins/whatsie.plugin/install.sh
89luca89: Do you think it's readable enough to put that URL in the Metadata?

Abhinav1217: No, Please don't put that url in Metadata. 
There is no Fedora version check and no Application version check. Refer to other plugins to see how we look for latest version in app urls.

Abhinav1217:
Facebook (current owner of Whatsapp)  does not appreciate 3rd party clients for whatsapp. They have their own web interface and a linux client is work in progress according to their blogs. 

Maybe a second opinion should be considered before adding this as even installer scripts can be considered a method of distribution.

89luca89:
Do you think it will create problems with FB Not wanting it to be released?
In the end it's only a wrapper over their website

But if it's risky indeed, feel free to freeze or close the PR

bddenhartog:
@Abhinav1217:

> Facebook (current owner of Whatsapp) does not appreciate 3rd party clients for whatsapp. They have their own web interface and a linux client is work in progress according to their blogs.

Can you link us to the sources you used for these two statements?

Abhinav1217:
**Whatsie has been abandoned and will not be developed further.** Original [github repo](https://github.com/Aluxian/Whatsie) is now 404. This is a backup repo https://github.com/gsantner/whatsie .  Maybe we should be looking for something which is still maintained. I am currently trying out http://meetfranz.com/

bddenhartog:
As @Abhinav1217 mentioned, the original repository for Whatsie is no longer available. The [backup repository][0] for Whatsie is deprecated. There is not an [official client][1] for Linux. As such, this PR is being closed without merging.

[0]: https://github.com/gsantner/whatsie
[1]: https://www.whatsapp.com/download/

Issue #466, opened 2017-07-17
[RFC] Add fedy to the RPM Fusion repo


kwizart:
As a RPM Fusion project coordinator I would like to invite you to consider to add fedy into the RPMFusion repository.

RPM Fusion is a community based repository and welcome anyone with good ideas to help to complement Fedora.

Please see the contribution guide (I can assist as a sponsor if you are not already a fedora packager)
https://rpmfusion.org/Contributors

The only point I would need to verify is if a particular snippet is causing conflict with RPM Fusion packages or repo, in which case it would be a matter to hide this behind a special "Enable with caution" menu.

What do you think ?

bddenhartog:
I think that, generally speaking, this is a great idea -- I do have some reservations, however, mostly around the delay in the release cycle due to the average review period while an update sits in `updates-testing` ([source](https://rpmfusion.org/Contributors#Updating_an_existing_package): _You'll have to wait a period ranging from 10 to 14 days for your package to be transferred to the "updates" repository._) This is, of course, very reasonable considering that this is not automated, and a real person is auditing the update, but it does delay our existing not-so-great deployment speed even further.

That said, after some review, refactoring, and implementation of proper update + deployment processes within the FWH organization, I'd be happy to look into submitting Fedy to the RPMFusion repository.

To give you (and other readers) a timeline, I'm currently looking at refactoring the organization's repositories to be easier to grok, and automate builds for -- in addition to standardizing Fedy's plugins and adding a better debugging experience (see https://github.com/folkswithhats/fedy/commit/0cb09c6f013b2908a6ac942845a230e665ffe3a4). Given my other workload and severe lack of free time, this will likely take me longer than I'd like, up to about a month or two from now. After that, I'll sit down and plan out a better cycle for updates, and _then_ look at RPMFusion.

On a side note, there is currently [an open issue](https://github.com/folkswithhats/fedy/issues/422) to refactor Fedy so that no "plugin" files are included, and instead pulled directly from github by the application, which would simply be a GUI browser for what we call "plugins". When this is complete, plugin updates would sneak past RPMFusion's review process (as the scripts would no longer be deployed with the gui, which is what users would be installing). Hopefully I've made this easy to understand, because I'd like to know how the RPMFusion organization would handle this - is that acceptable? It does open up the possibility that somehow a malicious script is merged into `master` (or really, a tagged release), which would not be seen by RPMFusion reviewers unless they browsed to the repository and inspected each plugin.

[...]

PR #531, opened 2017-09-01
Wiredesktop

89luca89:
Add Wire Desktop Client Application


plugins/wiredesktop.plugin/install.sh
```diff
@@ -0,0 +1,29 @@
...
+dnf -y install libappindicator-gtk*
+
+CACHEDIR="/var/cache/fedy/wire-desktop";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+pkgver=
+arch | grep x86_64
+if (( $? == 0));then
```
kwizart: you need a better test.

plugins/wiredesktop.plugin/install.sh
```diff
@@ -0,0 +1,29 @@
...
+	# 64 bit
+	URL=$(wget "http://mirror.yandex.ru/fedora/russianfedora/russianfedora/free/fedora/updates/25/x86_64/" -O - | grep wire-desktop |  sed 's/.*\">//g' | sed 's/<\/a.*//g' | tail -n 1)
```
kwizart: It's a very broken design for an installed.


Please consider asking the wire-desktop package maintainer to submit a request to RPM Fusion directly instead.

Abhinav1217: @kwizart  As fedy do provide packages which are yet to be available on Fusion repos, There is no need to  tell contributors to tell package maintainer something, unless contributor themselves are package maintainer.  

@89luca89 He is right, Please use Fedora release check. RussionFedora does try to keep up with fedora release cycle. Hardcoding release version will break dependencies.

kwizart: I totally disagree. A contribution, even if it looks "clean", or "working", can also be right or "wrong" based on that it's not the appropriate way to contribute. 
If this single package is introduced even if it could end in any repository, then the wire-desktop maintainer will lose the ability to issue a package update, fixing end-users bugs. 
At the end, it will remotely affect fedy, because it will be considered as a collection of bad hacks, and as it will bypass the relationship between end-users and maintainers.

Abhinav1217: @kwizart I did not say you are wrong about the plugin. Read my full comment. I have asked him to implement proper version check in his plugin. This plugin, at this state, is unacceptable.  I was referring to  the part where you were asking him to ask maintainer to submit to fusion repos directly. Maybe I should have phrased it better to avoid confusion.

89luca89: Ok so to be clear it's preferred to contact the Wire maintainer in russian-fedora and convince him to submit to rpmfusion?

plugins/wiredesktop.plugin/install.sh
bddenhartog: @kwizart When reviewing, please try to offer useful information - something like "you need to do this differently" or "you need a better test" doesn't help the contributor to understand _why_ a change is being requested, or how it _should_ be done (and why).

@89luca89, as in other reviews I've left, you can simply do:

```
if command arch | command grep 'x86_64'; then
   ...
fi
```

This helps the code be "self-documenting", a trait which is never bad :)

plugins/wiredesktop.plugin/install.sh
bddenhartog: @kwizart:

> It's a very broken design for an installer (then the fedora release check is non-existant).

Again, try to be clear when addressing _what_ the issue is with the code, and how the contributor can approach it differently.

> Please consider asking the wire-desktop package maintainer to submit a request to RPM Fusion directly instead.

As far as I can tell, `wire-desktop` does not currently exist in any of RPMFusion's repositories. While you and I have [spoken about moving Fedy to RPMFusion](https://github.com/folkswithhats/fedy/issues/466), that is not a move that has been made yet. Additionally, it is inappropriate (and offensive to both the contributor and the third party package maintainer) to request that a contributor to Fedy remove their contribution and instead ask an external package maintainer -- who is not involved in any of the aforementioned projects -- to submit their package to your platform's repositories. If you think that the third party package maintainer should submit their work to RPMFusion's repositories, please contact them directly instead of instructing _our_ contributors to do your work for you.

@89luca89 

> Ok so to be clear it's preferred to contact the Wire maintainer in russian-fedora and convince him to submit to rpmfusion?

No. Fedy's current methodology is okay with downloading files directly, as you've done here -- and as you've likely seen elsewhere in Fedy's plugins. However, @kwizart is correct in saying that this is a weak and unstable approach - if there's any change at all in the URI (the dir paths, package name, etc.) then the plugin breaks immediately for all users. This is something that [I have addressed](https://github.com/folkswithhats/fedy/issues/448) in the past.

All that said, it looks like RussianFedora actually provides [their own repositories](http://ru.fedoracommunity.org/stories/repository/), similar to RPMFusion, from which [this package](https://github.com/RussianFedora/wire-desktop) could be installed directly. Is there a reason that you have decided to bypass that?

89luca89:
I was thinking that being the RF Repo similar to RPMFusion, it can cause some conflict
So My idea was to simply pull the rpm and then let DNF sort the dependencies

Would it be a (bad) idea to install the repo, install the app and then disable the repo?

Suggestions are welcome of course, thanks

bddenhartog:
> I was thinking that being the RF Repo similar to RPMFusion, it can cause some conflict

`dnf` might require users to specify a repository when an identical package name is found in more than one repository; I'm actually not sure about this behavior (this should be tested). At a minimum, users could set the `cost` attribute of a repository to make it higher than the other, causing `dnf` to prefer the lower-cost repository ([source](http://dnf.readthedocs.io/en/latest/conf_ref.html#repo-options)).

> Would it be a (bad) idea to install the repo, install the app and then disable the repo?

This would prevent users from being able to upgrade the package via `dnf`, which is probably something we want to avoid.

89luca89:
@bddenhartog said:

> dnf might require users to specify a repository when an identical package name is found in more than one repository; I'm actually not sure about this behavior (this should be tested). At a minimum, users could set the cost attribute of a repository to make it higher than the other, causing dnf to prefer the lower-cost repository (source)

Isn't it a bit confusing for the user to have to manage repo costs?

bddenhartog:
@89luca89 said:

> Isn't it a bit confusing for the user to have to manage repo costs?

Yeah, it is -- I'm not sure there's a single 'best' solution here.

Abhinav1217:
Russian Fedora is [listed as a compatible] repository at fusion website.(https://rpmfusion.org/FedoraThirdPartyRepos),  [Their FTP](http://mirror.yandex.ru/fedora/russianfedora/russianfedora/free/fedora/updates/26/x86_64/) shows that they are including fusion repos, which suggests they are reusing any components already available in fusion. So there should be no conflicts if using it.  
Wire-desktop is not in F26 repos but in F25 repo, there are binaries for both 32 bit and 64 bit, Here, @89luca89  is downloading them using a very bad approch ( head or first link for 64 bit and tail for 32 bit,  badly assuming that the order of grep output would never change), He should instead look for `i686.rpm` for 32 bit and so on. Also even though i686 rpm is available in `x86_64` location, there is a dedicated ftp location `i386` for 32 bit binaries which could be used.   

**Alternate solution**
I don't know much about koji , Russian Fedora also have a koji system just like fedora have, and [they have listed F26 rpm for wire-desktop](http://koji.russianfedora.pro/koji/packageinfo?packageID=136), This also can be used. I have checked, it installs without any error and works fine. Thing to note is that it is also outdated build.

**Another alternate solution**
Since wire-desktop in RussionFedora repo is outdated, why don't we try to download official `.appimage` file from their website.  *Worst case scenerio,* both appimage and RuF method will disrupt users ability to update software using dnf. *Best case scenerio,* appimage does have an inbuilt update mechanism which app developer can use to update their app ( even has a delta update mechanism ).  There hasn't been any update since I started using the appimage, so I can't confirm if they are providing updates in appimage or not.  At present, the biggest issue using appimage in plugin is that gnome-wayland does not play well at launching GUI apps via terminal. A method I personally use is using `xhost +` and `xhost -` to temporarily provide permission to the script. But I don't thing this approach is something fedy should distribute, as this makes wayland open and insecure.  Alternately, we can choose a different download location instead of `/opt/`, something like `~/.config/fedy/appimage/wire-desktop.appimage` which would not require script to execute with `run-as-root` option (tested, used by me for etcher plugin). Since many publishers have started distributing .appimage, we can use that location for other plugins too.

**Another unofficial solution**
The [github discussion](https://github.com/wireapp/wire-desktop/issues/67) on rpm shows few users built .spec file and rpm repo, which can be used ( taking a leap of faith  with stranger ).  The one which looks most reliable is [hosted at gitlab](https://conoria.gitlab.io/wire-builds/). The [Spec file](https://gitlab.com/ConorIA/wire-builds) is also located there. I have tested it and it works fine. 

**_Sidenote_**
1. Someone can use the spec  file to create and maintain copr repo for wire-desktop. There is at present one copr repo for wire and it does not have a successful build.
2. Submit a package request to rpmfusion for wire-desktop and hope for best. Meanwhile use one of above methods in fedy.

bddenhartog:
@Abhinav1217 Thanks for looking into it. Just a couple of talking points...

> Worst case scenario ... RuF method will disrupt users ability to update software using dnf.

I assume you're talking about _other_ software here, as in, adding the repository would break the ability to upgrade a system. Given that it's listed as a compatible repository, is that really something that we should assume is possible?

> **Another unofficial solution**
>The github discussion on rpm shows few users built .spec file and rpm repo, which can be used (taking a leap of faith with stranger ). The one which looks most reliable is hosted at gitlab. The Spec file is also located there. I have tested it and it works fine.

At this point, why don't we simply use our own specfile and repository?

kwizart:
Unfortunately RussianFedora isn't compatible with Fedora anymore since it replaces some packages provided there. I will drop a note from the quoted page on rpmfusion.org.

I don't understand why you are twisting your head. This package should be provided by RPM Fusion.

Abhinav1217:
@bddenhartog 

> I assume you're talking about other software here,  

I was talking about wiredesktop. If we install wiredesktop by downloading rpm package from RuF instead of installing RuF repo, then user would not be able to update wiredesktop via dnf.  

>  why don't we simply use our own specfile and repository? That absolves us of any need for third-party trust, and we're already set up to serve RPMs.

In the sidenote, I have suggested we can do this via copr. I have thought about hosting our own package, but given that there several repository already dedicated for this purpose, like fusionrepos, negativo, and recently launched United-RPMs, Why create our own service when we could contribute to them.  https://imgs.xkcd.com/comics/standards.png

@kwizart 
>I don't understand why you are twisting your head. This package should be provided by RPM Fusion.

I have suggested that in sidenote, Wiredesktop clients are GPL3 open source, Server is also AGPL open source, I think fusion community would not have any issue providing it in fusion repos. 

>Unfortunately RussianFedora isn't compatible with Fedora anymore since it replaces some packages provided there. I will drop a note from the quoted page on rpmfusion.org.
 
I will test this some more. But While testing for wire-desktop, I have installed RuF repos on test VMs and today I ran system updates, I didn't face any issues. Even though the RuF26 FTP shows only few apps, RuF Koji system has almost all the apps listed in RuF25 FTP. That is why I have suggested @89luca89  that he can use Koji as download location.

PR #532, opened 2017-09-01
[WIP] Expand Software Offering: XMind

89luca89:
Add XMind Application


plugins/xmind.plugin/install.sh
```diff
@@ -0,0 +1,44 @@
...
+URL="http://dl2.xmind.net/xmind-downloads/xmind-8-update3-linux.zip"
+wget -c "$URL" -O "xmind.zip"
+if [[ ! -f "xmind.zip" ]]; then
+	exit 1
+fi
+rm -rf "/opt/xmind"
+mkdir -p "/opt/xmind"
+unzip -xq "xmind.zip" -d "/opt/xmind"
+
+if (( $? == 0));then
```
kwizart: I don't think this test is accurate to discover the right arch

plugins/xmind.plugin/install.sh
```diff
@@ -0,0 +1,44 @@
+CACHEDIR="/var/cache/fedy/xmind";
+
+mkdir -p "$CACHEDIR"
+cd "$CACHEDIR"
+
+arch | grep x86_64
```
kwizart: what's the point of that here ? you aren't making use of the test.

89luca89: Yeah I will fix this, It's not working this way

Should work on another method to pick the right arch?

bddenhartog:
Closing - rebase necessary and I was unable to push LFS data to the original author's repository. See #580.

PR #536, opened 2017-09-16
Fix dnf install

ArvindNair001:


bddenhartog:
Thanks! 

Did you run into an issue with the position of the `-y` parameter? This is actually completely valid in all of the versions of Fedora that we support, and while it is not my preferred code style, I don't see a reason to have a PR for such a minor formatting adjustment. It would make sense to include this in something along the lines of a repository-wide clean-up task.

ArvindNair001:
Yes I've tested your original code, as it is the only reason I've reached this repo because it had been aborting the install for some odd reason. I've never tried passing -y parameter before install but i think logically it makes sense for -y to be added after "dnf install" anyways the code worked for me after correction. I'm on fedora 26 btw !! and i love fedy  its a great initiative

bddenhartog:
Thanks for the kind words! I'm glad you find Fedy useful :)

> I've never tried passing -y parameter before install but i think logically it makes sense for -y to be added after "dnf install" anyways the code worked for me after correction

While the script may have worked for you after you made a change, the change in this pull request was not the reason for the failure. Passing the options before arguments is actually the correct format, as you can see from `man dnf`:

```
DNF(8)

NAME
       dnf - DNF Command Reference

SYNOPSIS
       dnf [options] <command> [<args>...]
```

bddenhartog:
P.S., I'd be more than happy to work with you to figure out what your issue was -- although I'm inclined to think it was probably just network-related.

PR #539, opened 2017-10-06
Fix install link

ArvindNair001:


bddenhartog:
This is neither a link nor an issue. You submitted this same changeset [here](https://github.com/folkswithhats/fedy/pull/536) (#536), to which I had the same reply:

`dnf -y install <package>` is perfectly valid. In fact, `dnf --help` shows that it prefers options _first_:

```
$ dnf --help
> usage: dnf [options] COMMAND
```

Did you run into an error with the `-y` option in the position it is in? I asked you (on the other thread) if you ran into a specific issue related to the position of the option, and you failed to answer me. Please open a terminal, run `fedy`, and paste any output you get from that moment until the moment that the sublime text plugin fails for you.

danstewart:
Closing - no response.

Issue #432, opened 2017-01-31
Add the "Adapta" theme by tista500


dralley:
https://github.com/adapta-project/adapta-gtk-theme

Optional wallpapers: https://github.com/adapta-project/adapta-backgrounds

COPR available here: https://copr.fedorainfracloud.org/coprs/heikoada/gtk-themes/

dralley:
Note that the COPR also includes the Arc and Paper themes

danstewart:
I wish I knew about the copr before I installed it manually yesterday.
I'll get this added tomorrow.

danstewart:
Looks like the copr is currently failing for adapta:
https://copr.fedorainfracloud.org/coprs/heikoada/gtk-themes/package/adapta-gtk-theme/

I'll try adding a plugin to add it manually.

danstewart:
@dralley Building manually takes a while so talking to the others about what we should do with this.
I've got a working install script if you want it though.
Gist: [adapta-gtk-theme-install.sh](https://gist.github.com/danstewart/179ca94277a3a6a4147b5709dec21602)

bddenhartog:
@danstewart any update on this?

danstewart:
@bddenhartog Not sure if I'm being an idiot but for some reason I can't install it from the [copr](https://copr.fedorainfracloud.org/coprs/heikoada/gtk-themes/):

```
$ sudo dnf copr enable heikoada/gtk-themes

You are about to enable a Copr repository. Please note that this
repository is not part of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Please do not file bug reports about these packages in Fedora
Bugzilla. In case of problems, contact the owner of this repository.

Do you want to continue? [y/N]: y
Repository successfully enabled.

$ sudo dnf install adapta-gtk-theme
Last metadata expiration check: 0:00:00 ago on Sat 28 Oct 2017 18:40:06 BST.
  * Maybe you meant: adapta-gtk-theme
No match for argument: adapta-gtk-theme
Error: Unable to find a match
```


I still have the manual script in a [gist](https://gist.github.com/danstewart/179ca94277a3a6a4147b5709dec21602) if you'd rather I just add that?

bddenhartog:
[...]

PR #542, opened 2017-10-29
Add adapta theme

danstewart:
#432 

Looks like the Cinnamon package conflicts with the arc icon theme we include so I've left it out for the time being.

### DIFF
--- a/plugins/adaptatheme.plugin/install.sh
+++ b/plugins/adaptatheme.plugin/install.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+dnf -y copr enable heikoada/gtk-themes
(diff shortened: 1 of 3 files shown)


Abhinav1217:
Does work as expected on Gnome. But I will need to do further testing on other spins. Because on KDE, It broke after restart.

bddenhartog:
@Abhinav1217 said:

> ...Because on KDE, It broke after restart.

Wouldn't that be a problem upstream? I don't see this PR doing any kind of custom scripting; it's just calling `dnf` to enable the COPR repository and install the packages -- something users would do themselves if they were so inclined. Since our current application architecture depends on each plugin's upstream source, I don't see an issue with accepting this is if tests fine under the default (Gnome-based) Fedora environment.

PR #545, opened 2017-11-08
Added papirus icon theme and Google Fonts

beli3ver:
Added papirus icon theme and Google Fonts


bddenhartog:
It doesn't look like the two plugins you're submitting in this pull request depend on one another. Please submit two different pull requests.

PR #547, opened 2017-11-09
Added Google Fonts

beli3ver:
The request again, seperated in two now :D


Abhinav1217:
Over 300 MB ( According to their readme). And Installs all the fonts available on google fonts. Some of the fonts ( Source code pro, source code sans, Ubuntu) are available from fedora repos and fusion repos directly. Contains fonts for other languages too.

bddenhartog:
@Abhinav1217 said:

> Do we require having over 300 MB worth of fonts. Some of the common fonts are installed by fedy itself. Many are available from fedora repos too.

I don't think the size of the archive matters. If a user wants all of the google fonts, they can get all of the google fonts -- it will simply take a while. It would be nice if we could display stuff like download size for a plugin, but our current application architecture doesn't allow for that.

I think that the key takeaway from the above comment, and reasons _not_ to accept this plugin, are:

* Fedy installs some of the fonts (source?)
* Many are available from the fedora repositories ([source](https://src.fedoraproject.org/))

I suppose what would really be a decision driver here is what is available through the official repositories, and what is not. I'd love it if someone were able to compile a table/list of this.

beli3ver:
I think a way to update the fonts is very difficult so a corp would be better.

Am 9. November 2017 18:29:32 MEZ schrieb Benjamin Denhartog <notifications@github.com>:
>@Abhinav1217 said:
>
>> Do we require having over 300 MB worth of fonts. Some of the common
>fonts are installed by fedy itself. Many are available from fedora
>repos too.
>
>I don't think the size of the archive matters. If a user wants all of
>the google fonts, they can get all of the google fonts -- it will
>simply take a while. It would be nice if we could display stuff like
>download size for a plugin, but our current application architecture
>doesn't allow for that.
>
>I think that the key takeaway from the above comment, and reasons _not_
>to accept this plugin, are:
>
>* Fedy installs some of the fonts (source?)
>* Many are available from the fedora repositories
>([source](https://src.fedoraproject.org/))
>
>I suppose what would really be a decision driver here is what is
>available through the official repositories, and what is not. I'd love
>it if someone were able to compile a table/list of this.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/folkswithhats/fedy/pull/547#issuecomment-343229684

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

Abhinav1217:
Size might not matter much in countries like US and Europe. But for example, Here in India, I have a 1GB per day plan, on mobile at 8mbps, and 60 GB per month Broadband at 4mpbs, (Download speed is approx 1/8th of plan). And as @beli3ver pointed out in next comment, Currently, if you are to update the package, it would mean over 300MB update even if the changes are done only on few fonts. Copr with delta package would be better. 

Also, Currently, Fedy doesn't show progress percentage, or launch terminal ( like the older version of fedy ), So it felt like the app has hung. I downloaded the file via command line directly, Wget reported length as unspecified during install, After completion, downloaded file was 364mb in size.

@bddenhartog I used to install a few fonts on new system from google fonts website, Namely, Noto, Roboto, Source Code Pro, and Ubuntu Fonts. Now all those fonts are available in Fedora repos itself. I did try to make a list of common fonts, but difference in naming convention means searching in dnf is not as straightforward as one might think. I also noticed some fonts are available in Google Fonts website but not on the github repo, Maybe there is some build process involved.   

_**Sidenote :**_

Just today on G+ feed I got a line on a new rust app called [fontfinder](https://github.com/mmstick/fontfinder). Its use is to preview google fonts, and install them in system. It is a new app with initial commit 2 days ago. But it is stable and works really well and blazingly fast. There is another app called [Typecatcher](https://github.com/andrewsomething/typecatcher/) which is older and stable ( FontFinder is rust implementation of It.) It to worked really well and was fast ( Not as fast as fontfinder, for some reason, Downloaded .ttf file were exactly same.

Instead of installing all the fonts available in Google Fonts, On users system, We can have plugin for such app, which would allow user to search for fonts they want and install it. Its kind of a middle ground solution.

plugins/googlefonts.plugin/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
Abhinav1217: Use cache dir `/tmp/fedy/<plugin-name> ` and download file there.

plugins/googlefonts.plugin/install.sh
```diff
@@ -0,0 +1,9 @@
...
+wget https://github.com/google/fonts/archive/master.tar.gz -O google-fonts.tar.gz
+tar -xf google-fonts.tar.gz
+mkdir -p /usr/share/fonts/truetype/google-fonts
+find $PWD/fonts-master/ -name "*.ttf" -exec sudo install -m644 {} /usr/share/fonts/truetype/google-fonts/ \; || return 1
+fc-cache -f > /dev/null
+rm -rf google-fonts.tar.gz fonts-master
```
Abhinav1217: Use Cache Dir in /tmp folder, No need to delete downloaded file.

bddenhartog:
@abhinav1217 brings up great points about preexisting solutions here. I'd be in favor of adding Typecatcher as a plugin to Fedy.

My comment regarding the download size wasn't in regards to "the data doesn't matter". I meant it more along the line of "it's not Fedy's responsibility to manage download sizes". I just wanted to be sure we were not rejecting a new, useful plugin due to a large download.

Anyhow, with all that said, I think closing this makes the most sense. It seems that we are in agreement that this particular plugin is not ideal in its current form. 

_P.S.: In general, let's not request changes on things that we haven't concluded acceptance for - no need to waste anybody's time!_

PR #561, opened 2017-11-28
Add Postman plugin

agc93:
I used JetBrains Toolbox as an example and haven't done a Fedy plugin before so let me know if there's any problems.

Current method results in the app installed at `/opt/Postman`, and the `.desktop` file installed as per other plugins. Have tested only on Fedora 26 (including clean install).

---

There seems to be some difference between plugins using `/var/cache/fedy/<app-name>` vs using `/tmp` for temp files. Is there a standard, or any docs?

### DIFF
--- a/plugins/postman.plugin/install.sh
+++ b/plugins/postman.plugin/install.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+CACHEDIR="/var/cache/fedy/postman";
+mkdir -p "$CACHEDIR"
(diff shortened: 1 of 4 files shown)


bddenhartog:
@agc93 Sorry for the delay in getting to this, been bogged down the past few months.

I'm kind of just going to blaze through PRs and issues today and tomorrow, get everything merged in and test from there before doing a release. That said, this looks to be fine. Any issues will be fixed post-merge by the Fedy team.

PR #563, opened 2017-12-03
Updated Sublime Text Version Status

nhubbard:
* Marked Sublime Text 2 as deprecated. Not officially deprecated, but no longer updated and new users will purchased Sublime Text 3 licenses that are not compatible with Sublime Text 2.
* Marked Sublime Text 3 as stable. Removing the `(beta)` tag from the description.

### DIFF
--- a/plugins/sublimetext2.plugin/metadata.json
+++ b/plugins/sublimetext2.plugin/metadata.json
@@ -1,7 +1,7 @@
 {
 	"icon": "sublime-text-2",
-	"label": "Sublime Text 2",
-	"description": "Sophisticated text editor for code, markup and prose. (deprecated)",
+	"label": "Sublime Text 2 (Deprecated)",
+	"description": "Sophisticated text editor for code, markup and prose.",
 	"proprietary": true,
 	"category": "Development Tools",
 	"scripts": {
--- a/plugins/sublimetext2.plugin/metadata.json
+++ b/plugins/sublimetext2.plugin/metadata.json
@@ -1,7 +1,7 @@
 {
 	"icon": "sublime-text-2",
 	"label": "Sublime Text 2",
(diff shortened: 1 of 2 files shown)


plugins/sublimetext2.plugin/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
 	"icon": "sublime-text-2",
 	"label": "Sublime Text 2",
-	"description": "Sophisticated text editor for code, markup and prose.",
+	"description": "Sophisticated text editor for code, markup and prose. (deprecated)",
```
bddenhartog: It would be better to append the deprecation notice to the `label` attribute.

Abhinav1217: Since Sublime text 3 has been [officially released](https://www.sublimetext.com/blog/articles/sublime-text-3-point-0), We can actually think about removing the st-2 plugin altogether.

Abhinav1217:
Since Sublime text 3 has been [officially released](https://www.sublimetext.com/blog/articles/sublime-text-3-point-0), We can actually think about removing the st-2 plugin altogether.

bddenhartog:
> Since Sublime text 3 has been officially released, We can actually think about removing the st-2 plugin altogether.

We should, however, retain a deprecated notice for at least one [minor version](https://semver.org/) so that users have a chance to remove it.

numacasa123:
:+1: bddenhartog

bddenhartog:
@nhubbard Are you going to have a chance to address the requested changes?

nhubbard:
@bddenhartog I'm so sorry :frowning: This was buried under all of my other GitHub notifications and I didn't get to see it. I'll get these fixed.

nhubbard:
@bddenhartog All fixed up. Looks good?

PR #564, opened 2017-12-09
Fix missing shared library libXss.so.1 on startup

adil452100:
Tested on a fresh f27 install, popcorn won't start because of a missing shared library libXss.so.1 provided by the libXScrnSaver package


SckyzO:
Hi all, 
New F27 install, need to install manually libXScrnSaver for Google Play Music. 

Regards,

PR #588, opened 2018-05-28
Android studio Icon Update

ArvindNair001:
Google has changed the icon name to studio.png


sudoforge:
Created an issue: #589

plugins/androidstudio.plugin/install.sh
```diff
@@ -21,7 +21,7 @@ unzip -xq "$FILE" -d "/opt/"
 
 ln -sf "/opt/android-studio/bin/studio.sh" "/usr/bin/android-studio"
 
-xdg-icon-resource install --novendor --size "scalable" "/opt/android-studio/bin/androidstudio.svg" "android-studio"
+xdg-icon-resource install --novendor --size 128 "/opt/android-studio/bin/studio.png" "android-studio"
```
sudoforge: It's unfortunate we no longer have a scalable svg, but oh well. Thanks for catching this.

Note to self: Even though the icon was renamed within the project to `studio.png`, it definitely makes sense to keep the system icon name as `android-studio`.

ArvindNair001: Great

PR #612, opened 2018-08-15
Adding Lutris plugin

dmanlfc:


plugins/lutris.plugin/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+#Get Fedora version
+Disty=$(lsb_release -i -s)
```
sudoforge: This variable name is unclear. We should prefer explicit variable names when possible, such as `distribution`.

However, since Fedy only supports Fedora, you might as well just hardcode `Fedora` into the URI.

plugins/lutris.plugin/install.sh
```diff
@@ -0,0 +1,11 @@
...
+Rel=$(lsb_release -r -s)
```
sudoforge: Using `lsb_release` is unsafe, as it might not exist on all Fedora installations (e.g. Docker, which is where tests are conducted).

You should instead use `rpm -E %fedora`. Additionally, please use a more appropriate variable name, e.g. `release`.

dmanlfc:
OK fixed - thanks for the suggestions

PR #622, opened 2018-10-06
Fix Simplenote install

### LINKED ISSUE #581: Simplenote Plugin issue
Fedy unable to install new versions of Simplenote ( only the old version )

Seems like Simplenote guys made a change in its filename pattern

In fedy script its appear as
 Simplenote-linux-x64.[0-9.]{5}.tar.gz
but the new filename pattern is
Simplenote-linux.[0-9.]{5}.tar.gz

After removing "-x64" seems like the new version is installed with Fedy.

Please verify.

danstewart:
Fixes #581 and adds images.


sudoforge:
@danstewart do you have time to knock the changes out today? I'll tackle them later if you don't.

danstewart:
@sudoforge Sorry, missed that last message. I've changed it over now.

It's a little clunky since the `tarball_url` is for the source tarball, not the release.

sudoforge:
> It's a little clunky since the tarball_url is for the source tarball, not the release.

I didn't check what was downloaded when PoCing it. You're right, this is downloading the full source code for the latest release. If you wanted to instead use the `.tar.gz` uploaded by the maintainers, you'd need to look at `assets[]` on the response:

```
TARBALL_URI=$(curl -s \
    -H "Accept: application/vnd.github.v3+json" \
    "${API_URI}/repos/automattic/simplenote-electron/releases/latest" |\
    jq -r '.assets[] | select(.name | contains(".tar.gz")) | .browser_download_url')
```

If you take this route, you should of course check to make sure that `TARBALL_URI` isn't empty. If `jq` doesn't match anything in `.assets[]` from the `select`, the result will simply be empty (the command won't fail).

danstewart:
@sudoforge Is it just something like this that you had in mind?

plugins/simplenote.plugin/install.sh
```diff
@@ -4,28 +4,49 @@ CACHEDIR="/var/cache/fedy/simplenote";
...
-URL="https://github.com/Automattic/simplenote-electron/$(wget https://github.com/Automattic/simplenote-electron/releases -O - | grep -Po releases/download/v[0-9.]{5}/Simplenote-linux-x64.[0-9.]{5}.tar.gz | head -n 1)"
-FILE=${URL##*/}
+API_URI="https://api.github.com"
+
+ASSETS=$(curl \
+    -H "Accept: application/vnd.github.v3+json" \
+    "${API_URI}/repos/automattic/simplenote-electron/releases/latest" |\
+    jq -cr '.assets[]')
+
+for asset in $ASSETS; do
```
sudoforge: This `for` loop is unecessary. See [my comment here][0].

[0]: https://github.com/fedy/fedy/pull/622#issuecomment-428321896

plugins/simplenote.plugin/.gitattributes
```diff
@@ -0,0 +1,2 @@
+simplenote.png filter=lfs diff=lfs merge=lfs -text
+128x128.png filter=lfs diff=lfs merge=lfs -text
```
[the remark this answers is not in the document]
sudoforge: This file still needs to be deleted.

plugins/simplenote.plugin/install.sh
```diff
@@ -4,28 +4,41 @@ CACHEDIR="/var/cache/fedy/simplenote";
...
 cd "$CACHEDIR"
 
-URL="https://github.com/Automattic/simplenote-electron/$(wget https://github.com/Automattic/simplenote-electron/releases -O - | grep -Po releases/download/v[0-9.]{5}/Simplenote-linux-x64.[0-9.]{5}.tar.gz | head -n 1)"
-FILE=${URL##*/}
+API_URI="https://api.github.com"
+
+URL=$(curl \
+    -H "Accept: application/vnd.github.v3+json" \
+    "${API_URI}/repos/automattic/simplenote-electron/releases/latest" |\
+    jq -r '.assets[] | select(.name | contains("x64.tar.gz")) | .browser_download_url')
```
sudoforge: Sorry for just noticing this, but they actually release RPMs. Is there a reason we're using the tarball and installing manually?

sudoforge: Also we should probably accomodate 32-bit systems too.

danstewart:
@sudoforge Looks like they only added RPMs yesterday, nice timing.

I've changed to fetch the RPM and (un)install via `dnf`.

PR #642, opened 2019-02-02
Add install from source instructions

danstewart:


README.md
```diff
@@ -30,6 +30,23 @@ Installation instructions can be found on https://www.folkswithhats.org.
...
 
+### Install from source
+
+```
+# RPMFusion
+sudo dnf install https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://download1.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
+
+# Install other dependencies - https://github.com/fedy/fedy#dependencies
+# eg.
+sudo dnf install gjs
```
sudoforge: If we're providing instructions for installation, we should provide the complete instructions. "Install other dependencies" isn't super useful and requires the reader to reference another list. We should instead remove the `[Dependencies]` section of the readme and document the deps here.

README.md
```diff
@@ -30,6 +30,23 @@ Installation instructions can be found on https://www.folkswithhats.org.
 
 _Do not follow instructions from any other source!_
 
+### Install from source
```
sudoforge: `Install` should instead be `Build`.

danstewart:
@sudoforge I've made the requested changes.

I think it's still worth having the separate list of dependencies outwith the build from source instructions.

sudoforge:
> I think it's still worth having the separate list of dependencies outwith the build from source instructions.

I'm against this because it introduces yet a _third_ place that dependencies need to be updated.

danstewart:
Sorry, didn't realise you changed the README after this PR.

Where is the 3rd place the dependencies would be listed - dependencies list, build instructions and?

My issue is I don't like them *only* being in the build instructions but if they are somewhere else then yeah duplicating them again seems unnecessary.

sudoforge:
The dependency list exists in the spec file, and in the build-from-source instructions. If we add another itemized list of deps, that makes a third place that will have to be updated when/if dependencies are updated. Maybe this isn't a problem, and I'm being silly.

I'm also not sure they belong anywhere _except_ in the build-from-source instructions. A user who isn't building from source shouldn't have to care about dependencies, right?

danstewart:
I see what you mean, fair point.
I've rebased my branch to include your changes from 0d2a30c53f459dc5f7864e2a18068c619bdb7824 which removes the dependencies section.

Issue #644, opened 2019-03-19
Dino icon image too large?


svenevs:
Just installed from source following directions here https://github.com/fedy/fedy/issues/632#issuecomment-466655124

![dino](https://user-images.githubusercontent.com/5871461/54626501-80b7f000-4a2e-11e9-9c0b-133a4dfa9cc1.png)

Should Dino be that big? I love the icon so it doesn't matter much to me.  I tried to look around but am not very familiar with how fedy works...happy to provide any testing with guidance, if fix is simple I'll send PR?

Thanks for Fedy, happy "customer" for years :slightly_smiling_face:

sudoforge:
Well, you can't expect a _dinosaur_ to be small and byte-sized, now can you? :)

Thanks for noticing this. I'll spin up a VM and attempt to recreate it as soon as I am able.

danstewart:
PR #652 resizes the icons until a better fix is implemented.

Looks like we only scale the icons if they exist on the icon pack you use - if we use the icon that's bundled with fedy we just display it as the actual size.

Unsure why this only happens when building from source though...

PR #652, opened 2019-05-25
Resize icons

danstewart:
Workaround for #644 

Issue as that if the icon does not exist in the icon pack we don't attempt to resize the image:
https://github.com/fedy/fedy/blob/master/app.js#L504-L520

This PR just resizes them for now but will look into a better solution at some point.


sudoforge:
I'll be around and working through the backlog today.

Could we get `refers 644` or `closes 644` in the commit body?

danstewart:
@sudoforge Commit message updated. 

Thought about it a little more and I'm not sure if it's temporary or not.
The issue is we don't automatically scale icons that we bundle with fedy, but I suppose we should be bundling correctly sized icons in the first place (48x48 in size).

I think we can scale them automatically when displaying (though we might need to add a dependency) but we may as well just scale them when we add them in the first place.

danstewart:
Just reverted the change to `.gitattributes` too.
I tracked all 4 of the icons with `git lfs` unnecessarily.

danstewart:
Nevermind, that's made a mess of it since it was tracked with `git lfs` and now it's not...
I'll redo this.

